PR #112, opened 2026-04-26
Fix GC rooting gap for loop completion values

pmatos:
### Motivation
- Loop backedge safepoints could trigger a GC while the loop completion value `v` was only held in a Rust local and not visible to the GC, allowing that object to be reclaimed and producing stale object IDs or panics when `v` was later used.
- The change ensures loop completion values are preserved across safepoints so that allocations inside tight loops cannot cause availability failures.

### Description
- Temporarily root the loop completion value by calling `gc_root_value(&v)` immediately before each `gc_safepoint()` and `gc_unroot_value(&v)` immediately after to keep `v` live across collections.
- Applied the temporary rooting to all loop execution paths in `src/interpreter/exec.rs`, including `exec_while`, `exec_do_while`, labeled loop handling, the `for` loop per-iteration path, `for...in` iteration, and `for...of`/`exec_for_of_loop`.
- This is a minimal remediation that preserves existing GC semantics while preventing collection of live loop-local values.

### Testing
- Ran the test suite with `cargo test -q` and observed all interpreter tests pass (`60 passed; 0 failed`).
- No automated test regressions were detected by the run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e7438a703083328563df281191a47c)

### DIFF
--- a/src/interpreter/exec.rs
+++ b/src/interpreter/exec.rs
@@ -1506,7 +1506,9 @@ impl Interpreter {
     fn exec_while(&mut self, w: &WhileStatement, env: &EnvRef) -> Completion {
         let mut v = JsValue::Undefined;
         loop {
+            self.gc_root_value(&v);
             self.gc_safepoint();
+            self.gc_unroot_value(&v);
             let test = match self.eval_expr(&w.test, env) {
                 Completion::Normal(v) => v,
                 other => return other,
@@ -1539,7 +1541,9 @@ impl Interpreter {
     fn exec_do_while(&mut self, dw: &DoWhileStatement, env: &EnvRef) -> Completion {
         let mut v = JsValue::Undefined;
         loop {
+            self.gc_root_value(&v);
             self.gc_safepoint();
+            self.gc_unroot_value(&v);
             match self.exec_statement(&dw.body, env) {
                 Completion::Normal(val) => {
                     v = val;
@@ -1574,7 +1578,9 @@ impl Interpreter {
             Statement::While(w) => {
                 let mut v = JsValue::Undefined;
                 loop {
+                    self.gc_root_value(&v);
                     self.gc_safepoint();
+                    self.gc_unroot_value(&v);
                     let test = match self.eval_expr(&w.test, env) {
                         Completion::Normal(v) => v,
                         other => return other,
@@ -1611,7 +1617,9 @@ impl Interpreter {
             Statement::DoWhile(dw) => {
                 let mut v = JsValue::Undefined;
                 loop {
+                    self.gc_root_value(&v);
                     self.gc_safepoint();
+                    self.gc_unroot_value(&v);
                     match self.exec_statement(&dw.body, env) {
                         Completion::Normal(val) => {
                             v = val;
@@ -1709,7 +1717,9 @@ impl Interpreter {
         let mut v = JsValue::Undefined;
         let result = 'for_loop: {
             loop {
+                self.gc_root_value(&v);
                 self.gc_safepoint();
+                self.gc_unroot_value(&v);
                 if let Some(test) = &f.test {
                     let val = match self.eval_expr(test, &iter_env) {
                         Completion::Normal(v) => v,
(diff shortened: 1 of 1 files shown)


pmatos (commented):
## Review summary

Targeted fix that wraps every loop-backedge `gc_safepoint()` in `exec.rs` with a paired `gc_root_value(&v)` / `gc_unroot_value(&v)` so the loop completion value held in the Rust local `v` cannot be reclaimed mid-collection. The mechanism (push to `gc_temp_roots`, walked from `collect_roots` in `gc.rs:89`) is the same one already used for iterators elsewhere (e.g. `eval.rs:6772`, `exec.rs:2041`), so the new root location is definitely visited by the marker. The diff is balanced (push then pop around exactly the safepoint window), avoids the call_function-style trap because `v` is rooted *during* the collection rather than *before* an arbitrary callee runs, and the pop happens unconditionally so abrupt completions cannot leak roots.

Loop-kind coverage is complete:
- `exec_while` (1506)
- `exec_do_while` (1539)
- `exec_labeled_loop` -> labeled `while` (1574) and labeled `do-while` (1611)
- `exec_for` (1652) -- labeled `for` reuses this via the dispatcher at line 917
- `exec_for_in` (1771) -- labeled `for-in` reuses this via line 918
- `exec_for_of_loop` (2047) -- labeled `for-of` reuses this via line 919

Performance impact is a single `Vec::push` plus an `rposition` + `Vec::remove` per iteration. Both touch a small vec (depth equals nesting of currently rooted temporaries, typically single digits), so the cost in hot loops is negligible.

Build sanity: only safe code, no lifetime gymnastics, no new `Rc::clone`s. Should compile cleanly.

## Findings

### [major] No regression test for the GC-pressure scenario
This is a memory-correctness fix that is invisible unless GC fires at exactly the wrong moment. Without a test that *forces* allocation pressure mid-loop while holding a heap-object completion value (e.g. a tight `while`/`for` whose body produces an object literal as its completion and allocates enough to hit `gc_threshold_bytes`), the bug can silently regress when a future GC change shifts safepoint placement or threshold tuning. Recommend adding a test under `test262-extra/` (or a Rust-level test exercising `gc_requested = true` then asserting the completion value's object id still resolves) for at least one loop kind.

### [major] Same vulnerability remains in non-loop safepoint sites
The PR only patches loop backedges, but the identical pattern exists at three other safepoint sites that hold a Completion value in a Rust local across the safepoint:
- `src/interpreter/exec.rs:112` -- `exec_statements`: `result` carries the previous statement's completion JsValue across `gc_safepoint()` at the top of the loop.
- `src/interpreter/eval.rs:12756` -- `eval`'s body loop: same pattern with `last`.
- `src/interpreter/eval/modules.rs:1120` -- `run_module`'s body loop: same pattern with `result`.

In typical code these completion values are bound through the environment so they are independently rooted, but the bug surface is identical to the loop case (an object-literal expression statement followed by a statement that triggers GC). Worth either fixing in this PR or filing a follow-up issue so they are not forgotten.

### [nit] Pre-existing related gap in `exec_for_in`
`obj_val` at `src/interpreter/exec.rs:1825` is held only as a Rust local across `gc_safepoint()` at line 1849, and the loop body uses `obj_id` (a bare `u64`) afterwards. If the iterable object is unreachable elsewhere (e.g. `for (let k in {a:1,b:2,c:3}) { ... }`), GC could reclaim that slot and `obj_id` would become dangling. Not introduced by this PR and out of scope for the title, but a good candidate for a follow-up: roll `obj_val` into the new rooting pattern alongside `v`.

### [nit] Consider an RAII guard
The `push`/`pop` pattern is repeated seven times. A small RAII guard (`let _g = self.gc_root_scope(&v); self.gc_safepoint();`) would shrink the diff at each site, eliminate the (currently impossible but easy-to-introduce) bug of forgetting the unroot on a new code path, and avoid the `rposition` scan in `gc_unroot_value` if implemented as a stack-discipline pop. Not a blocker for this PR.

## Verdict
LGTM as a targeted fix for the loop-backedge case. Strongly recommend adding a forcing-GC regression test before merge so this cannot silently regress, and filing a follow-up for the matching pattern in `exec_statements`, `eval`, and `run_module`.

PR #113, opened 2026-04-26
Harden Claude workflow against untrusted comment triggers

pmatos:
### Motivation
- The `claude` GitHub Actions workflow previously ran whenever a comment/issue contained `@claude` and granted sensitive `id-token: write` permission and a repository secret to a third‑party action, allowing untrusted users to trigger privileged execution.

### Description
- Restrict the `jobs.claude.if` expression in `.github/workflows/claude.yml` to require the comment/issue author's `author_association` be one of `OWNER`, `MEMBER`, or `COLLABORATOR` in addition to the existing `@claude` mention checks for `issue_comment`, `pull_request_review_comment`, `pull_request_review`, and `issues` events.
- Preserve existing behavior and permissions while adding the trust gating logic using `contains(fromJSON(...), ...)` checks so only trusted actors can trigger the job.

### Testing
- Ran `git diff --check` to verify no formatting/errors were introduced and it succeeded.
- Inspected the workflow diff with `git diff -- .github/workflows/claude.yml` to confirm the updated `if` expression was applied as intended.
- Listed the updated workflow file with `nl -ba .github/workflows/claude.yml` to validate the final contents.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69b56da37b788332813fd2cac84138ca)

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -13,10 +13,10 @@ on:
 jobs:
   claude:
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude') && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude') && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude') && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.review.author_association)) ||
+      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')) && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association))
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -47,4 +47,3 @@ jobs:
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
           # claude_args: '--allowed-tools Bash(gh pr:*)'
-


pmatos (commented):
## Security review: Claude workflow trust gating

### Summary
This PR adds `author_association` gating to the four trigger branches in `.github/workflows/claude.yml`, requiring the actor to be `OWNER`, `MEMBER`, or `COLLABORATOR` on top of the existing `@claude` mention check. That closes the most obvious attack vector: any unauthenticated GitHub user opening an issue or commenting on a PR with `@claude` could previously trigger a privileged workflow that has `id-token: write` and consumes `secrets.CLAUDE_CODE_OAUTH_TOKEN`.

I walked the diff against a malicious-fork-PR threat model and the gate holds. **Recommending approval**, with a couple of non-blocking notes for future hardening.

### Threat model walkthrough (what changes for an attacker)

Pre-PR: a fork user opens a PR, then comments `@claude please help` from their own (untrusted) account. The `issue_comment` branch fires, the job runs, the Anthropic OAuth secret is mounted into the runner environment, the Claude action reads the comment body as instructions, and a prompt-injected comment can attempt to exfiltrate the secret or otherwise abuse the agent. The first-time-contributor approval gate that GitHub applies for `pull_request` does **not** apply to `issue_comment` events on existing PRs.

Post-PR: the same comment now fails the `if:` because `github.event.comment.author_association` is `NONE` (or `FIRST_TIME_CONTRIBUTOR`/`CONTRIBUTOR`), neither of which is in the allowlist. The job never starts, the secret is never minted into env, prompt-injection surface is gone. Verified for all four event types.

### Findings

**[major - non-blocking] `COLLABORATOR` includes Triage and Read roles** — `.github/workflows/claude.yml:15-19`
GitHub's `author_association` returns `COLLABORATOR` for any user with collaborator access at any level (Read, Triage, Write, Maintain, Admin). A user added with Triage permission can comment but cannot push code, yet they would pass this gate and trigger the privileged workflow. For a personal repo with one collaborator this is fine; if collaborators are added with limited roles in the future, the gate is laxer than it appears. If you want belt-and-braces, gate at the action layer too (e.g., `anthropics/claude-code-action` honours an actor allowlist), or move to a hardcoded `github.actor == 'pmatos'` check until the contributor base grows. Not a blocker today.

**[nit] No threat-model comment in the workflow** — `.github/workflows/claude.yml:15`
The four-line `if:` is the entire load-bearing security boundary. A future edit (e.g., adding a fifth trigger, or someone "simplifying" the boolean) can silently drop the `author_association` check. A short comment block above the `if:` referencing this PR / the threat (untrusted commenter triggering a workflow that consumes a secret) would help. PR description has the rationale but the file does not.

**[nit] `issues: [opened, assigned]` semantics changed slightly**
For `issues.assigned` the new gate checks `github.event.issue.author_association` (the *issue opener*'s association), not the *assigner*'s. So a maintainer assigning themselves to an issue opened by an external user with `@claude` in the body will no longer trigger Claude. That's strictly a security improvement (the previous behaviour let any external user seed a `@claude` payload that fired when a maintainer touched the issue), but worth knowing. Maintainers can re-trigger by just commenting `@claude ...` themselves.

**[nit] `claude-code-review.yml` is a different shape and out of scope here**
Sibling workflow `claude-code-review.yml` fires on `pull_request` (not `pull_request_target`), so the `GITHUB_TOKEN` is read-only and GitHub's first-time-contributor approval gate covers the secret. It doesn't need the same change. Flagging only because reviewers may wonder why the fix is one-sided.

### Things I checked and found clean

- No `${{ github.event.*.body }}` (or other user-controlled field) is interpolated into a `run:` block, so no shell-injection vector is introduced or pre-existing.
- `fromJSON('["OWNER","MEMBER","COLLABORATOR"]')` is properly quoted; the surrounding `'` and inner `"` parse correctly as a YAML scalar and a JSON array.
- Workflow uses `actions/checkout@v6` pinned by major; PR does not touch this.
- `permissions:` block already drops `contents` to `read`. No widening introduced.
- Trigger list is **not** `pull_request_target`, so attacker-controlled fork code is never checked out alongside secrets even on the legitimate path.
- `id-token: write` is only used by `claude-code-action` for OIDC to Anthropic, not for repo writes.
- Diff is exactly one file, one commit, no stray changes.

### Verdict
Approve with the two nits above as follow-ups. The hardening is correct, minimal, and closes the realistic attack. The remaining `COLLABORATOR` breadth is acceptable for a single-maintainer repo and easy to tighten later if needed.

pmatos:
Thanks for the thorough security review. Triaged by `pm-autofix-pr` after dual-evaluator agreement between Claude Opus 4.7 and Codex CLI — disposition per finding below.

### 1. `[major - non-blocking]` COLLABORATOR breadth (lines 15-19)

**Deferred (diminishing returns).** Both evaluators agreed this is a real future-hardening concern but not in-PR work, since the repo currently has no external collaborators added with limited roles and the PR's stated scope is closing the unauthenticated-trigger vector. Tracked as #138 with three concrete options (drop `COLLABORATOR`, literal `github.actor` check, or layer the action's own actor allowlist on top).

### 2. `[nit]` Threat-model comment above the `if:` (line 15)

**Deferred for separate discussion.** The two evaluators split: Claude Opus 4.7 said REJECT on style-preference grounds (CLAUDE.md says "Use comments sparingly. Only comment complex code."), Codex CLI said DEFER on scope-creep grounds. Per the dual-evaluator decision matrix any disagreement defaults to DEFER, so this is tracked as #139 with illustrative wording — to be revisited when the maintainer decides whether the workflow-security boundary is "complex code" worth annotating.

### 3. `[nit]` `issues.assigned` semantics change

**No action taken.** Both evaluators agreed this is informational, not actionable — the new behavior is the desired behavior ("strictly a security improvement"), and the workaround (a maintainer comments `@claude …` directly) is trivial. Nothing to fix or track.

### 4. `[nit]` `claude-code-review.yml` is out of scope

**No action taken.** Both evaluators agreed — the reviewer themselves flagged this as out of scope and explained why the sibling workflow does not need the same change (uses `pull_request`, not `pull_request_target`; `GITHUB_TOKEN` is read-only; GitHub's first-time-contributor approval gate already covers the secret).

### CI note

The `claude` check on this PR fails with *"Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch."* This is the action's intentional self-protection against PRs that modify the workflow file itself, and the action explicitly states *"this is normal and you should ignore this error."* It will pass once merged.

---

_This assessment was made by two independent AI reviewers (Claude Opus 4.7 and Codex CLI). If you disagree with any disposition, please reply and we'll re-evaluate._

PR #115, opened 2026-04-26
Bound proxy ownKeys trap result length

pmatos:
### Motivation
- The proxy `ownKeys` invariant code trusted the trap result's `length` and directly allocated/iterated `0..length`, allowing a malicious proxy to request huge allocations and trigger OOM/DoS.
- The change aims to mitigate the availability risk by bounding work done for attacker-controlled `ownKeys` results while preserving normal behavior for reasonable key counts.

### Description
- Add a hard upper bound `MAX_PROXY_OWNKEYS_RESULT_LEN = 1_000_000` and apply it before converting the trap `length` into a `usize` to prevent unbounded allocations.
- Validate `length` with `is_finite()` and `n > 0.0` and use `n.floor() as usize` when converting to `usize` to follow `ToLength`-style semantics.
- Enforce the bound in three places: `validate_ownkeys_invariant`, `proxy_own_keys` (the main `[[OwnPropertyKeys]]` proxy path), and `enumerable_own_keys` in `helpers.rs` so all proxy-driven enumeration paths are protected.
- Return a `TypeError` when the trap `length` exceeds the supported limit or is not a valid numeric length, preserving existing error responses for invalid results.

### Testing
- Ran `cargo fmt -- src/interpreter/eval.rs src/interpreter/helpers.rs` which completed successfully.
- Ran `timeout 180 cargo check -q`, which completed successfully and emitted existing `dead_code` warnings; no new compile errors were introduced.
- No new unit tests were added; changes are minimal and focused on input validation to mitigate the reported DoS vector.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69b34bc5ef348332872198020ec0d035)

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -14074,9 +14074,19 @@ impl Interpreter {
         trap_result: &JsValue,
         target_val: &JsValue,
     ) -> Result<(), JsValue> {
+        const MAX_PROXY_OWNKEYS_RESULT_LEN: usize = 1_000_000;
+
         let trap_keys: Vec<String> = if let JsValue::Object(arr) = trap_result {
             let len = match self.get_property_on_id(arr.id, "length") {
-                JsValue::Number(n) => n as usize,
+                JsValue::Number(n) if n.is_finite() && n > 0.0 => {
+                    let len = n.floor() as usize;
+                    if len > MAX_PROXY_OWNKEYS_RESULT_LEN {
+                        return Err(self.create_type_error(
+                            "'ownKeys' on proxy: trap result length exceeds supported limit",
+                        ));
+                    }
+                    len
+                }
                 _ => 0,
             };
             let arr_id = arr.id;
@@ -15368,6 +15378,8 @@ impl Interpreter {
     /// Proxy-aware [[OwnPropertyKeys]] - checks proxy `ownKeys` trap, recurses on target if no trap.
     /// Returns all own property keys (for getOwnPropertyNames).
     pub(crate) fn proxy_own_keys(&mut self, obj_id: u64) -> Result<Vec<JsValue>, JsValue> {
+        const MAX_PROXY_OWNKEYS_RESULT_LEN: usize = 1_000_000;
+
         if self.get_proxy_info(obj_id).is_some() {
             let target_val = self.get_proxy_target_val(obj_id);
             match self.invoke_proxy_trap(obj_id, "ownKeys", vec![target_val.clone()]) {
@@ -15386,7 +15398,16 @@ impl Interpreter {
                             _ => JsValue::Undefined,
                         };
                         let len = match len_val {
-                            JsValue::Number(n) => n as usize,
+                            JsValue::Number(n) if n.is_finite() && n > 0.0 => {
+                                let len = n.floor() as usize;
+                                if len > MAX_PROXY_OWNKEYS_RESULT_LEN {
+                                    return Err(self.create_type_error(
+                                        "'ownKeys' on proxy: trap result length exceeds supported limit",
+                                    ));
+                                }
+                                len
+                            }
+                            JsValue::Number(_) => 0,
                             _ => {
                                 return Err(self.create_type_error(
                                     "ownKeys trap result length is not a number",
--- a/src/interpreter/helpers.rs
+++ b/src/interpreter/helpers.rs
@@ -309,8 +309,17 @@ pub(crate) fn enumerable_own_keys(
                     interp.validate_ownkeys_invariant(&v, &target_val)?;
                     let mut keys = Vec::new();
                     if let JsValue::Object(arr) = &v {
+                        const MAX_PROXY_OWNKEYS_RESULT_LEN: usize = 1_000_000;
                         let len = match interp.get_property_on_id(arr.id, "length") {
-                            JsValue::Number(n) => n as usize,
+                            JsValue::Number(n) if n.is_finite() && n > 0.0 => {
+                                let len = n.floor() as usize;
+                                if len > MAX_PROXY_OWNKEYS_RESULT_LEN {
+                                    return Err(interp.create_type_error(
+                                        "'ownKeys' on proxy: trap result length exceeds supported limit",
+                                    ));
+                                }
+                                len
+                            }
                             _ => 0,
                         };
                         for i in 0..len {


pmatos (commented):
## Summary

This PR bounds the `ownKeys` proxy trap result length at 1,000,000 in three call sites, throwing a `TypeError` when exceeded. The cap is correctly placed *before* allocation/iteration in each site, the error type is correct (matches existing proxy invariant violations), and the addition of `is_finite()` / `n > 0.0` / `n.floor()` guards plugs the `Infinity`-as-`usize::MAX` DoS that was the worst-case path.

I traced the three sites in `src/interpreter/eval.rs` (`validate_ownkeys_invariant` @ ~14068, `proxy_own_keys` @ ~15378) and `src/interpreter/helpers.rs` (`enumerable_own_keys` @ ~314), and the reasoning is sound: even though `enumerable_own_keys` calls `validate_ownkeys_invariant` *before* the local cap check, the cap that the patch adds inside `validate_ownkeys_invariant` itself stops the unbounded iteration at the outermost layer. So the DoS is shut down before any `0..len` walk. Spec divergence is intentional (ECMA-262 §10.5.11 imposes no cap) and mirrors V8/SpiderMonkey-style hardening.

**IMPORTANT — duplicate PR:** PR #116 ("Bound proxy ownKeys trap result length") is **byte-identical** to this one — same diff, same description, same author, same Codex task; only the branch SHA differs. Both were generated by the codex agent. Recommend closing one without merging. Either is fine to take; no preference on technical grounds.

## Findings

### Major

- **[major] No regression test for the DoS scenario or for the bound's edge.** `src/interpreter/eval.rs:14068`, `15378`, `helpers.rs:314`. Per CLAUDE.md ("Any validation that's spec-correct but not in test262 should have its own tests in test262-extra"), this PR ships a deliberate non-spec hardening with zero tests. Please add at least:
  - `test262-extra/...proxy-ownkeys-length-bound-throws.js`: `new Proxy({}, { ownKeys: () => ({ length: 1_000_001 }) })` reflected via `Reflect.ownKeys` / `Object.keys` / `Object.getOwnPropertyNames` — must throw `TypeError`, must complete in <1s.
  - A passing companion at the cap (e.g. `length: 1_000_000` returning a real array of strings) to lock in that legitimate use is *not* rejected.
  - A fast-fail variant for `length: Infinity` and `length: 2**32 - 1` (the actual exploit vector cited in the threat model). Without these, future refactoring can silently break the cap and there's no signal in CI.

- **[major] `Infinity` and negative `length` are silently coerced to 0 instead of throwing.** `src/interpreter/eval.rs:14068-14077`, `15870-15879`; `src/interpreter/helpers.rs:314-322`. The new guard `JsValue::Number(n) if n.is_finite() && n > 0.0` falls through to `_ => 0` for `Infinity`, `NaN`, `-0`, and negative numbers. For `NaN`/`-N` this matches prior behavior (saturating cast was 0) — fine. But for **`Infinity` the prior behavior was the DoS itself** (`Infinity as usize` → `usize::MAX`), which is exactly the attack vector the PR is fixing. Silently treating it as 0 (i.e., "this proxy has no keys") rather than throwing creates an asymmetry: a trap returning `length: 1_000_001` throws, but `length: Infinity` returns `[]`. Consider folding `!n.is_finite() || n < 0.0` into the same TypeError branch (or at minimum throw on `Infinity`), so the cap is the *only* exit for "length too big."

- **[major] Inconsistency between the three sites for the non-Number-length case.** In `proxy_own_keys` the original code throws on non-numeric `length` ("ownKeys trap result length is not a number"). In `validate_ownkeys_invariant` and `enumerable_own_keys` the original code silently uses 0. The patch preserves this inconsistency. Not a regression, but if the goal is "harden ownKeys handling," the three sites should share semantics. A single helper `fn validated_ownkeys_length(&mut self, len_val: &JsValue) -> Result<usize, JsValue>` (returning `Err(TypeError)` for both `>1M` and non-Number/Infinity/negative) would deduplicate the cap constant, the floor logic, and the validation across all three callers.

### Minor

- **[nit] `MAX_PROXY_OWNKEYS_RESULT_LEN` is redeclared inside three functions.** `src/interpreter/eval.rs:14064`, `15852`, `src/interpreter/helpers.rs:560`. Hoist to a module-level `const` (ideally next to other proxy constants in `eval.rs`) so the cap can be tuned in one place and isn't easy to drift apart accidentally. This also makes test262-extra tests easier to write referentially.

- **[nit] Error message could be more actionable.** `"trap result length exceeds supported limit"` doesn't tell users what the limit is. Compare to V8: `"Proxy ownKeys returned more than 2^25 keys"`. Consider `"'ownKeys' on proxy: trap result length {n} exceeds engine limit of {MAX}"` so debugging is easier.

- **[nit] `Vec::with_capacity(len)` after the cap is fine but could OOM under memory pressure for ~1M-entry proxies.** `src/interpreter/eval.rs:15387`. Not a regression — just noting that 1M entries × `JsValue` size is tens of MB. Acceptable for the cap value chosen.

### Spec / sibling-vector check (informational, not blocking this PR)

- The PR scope is correctly limited to `ownKeys`. Other proxy trap return values (`get`, `set`, `has`, `defineProperty`, `getOwnPropertyDescriptor`, `getPrototypeOf`) return scalars/objects/booleans, not array-likes, so they don't have the same vector. The `getOwnPropertyDescriptor` trap return is iterated for property names but only over a fixed small set (`value`, `writable`, `enumerable`, `configurable`, `get`, `set`), not attacker-controlled length.
- Non-proxy targets with pathological own-key counts (e.g. a sparse `Array(2**32-1)` passed as the proxy target with no `ownKeys` trap) fall through to `OrdinaryOwnPropertyKeys`, which iterates `property_order` — that's the engine's own bookkeeping, not attacker-supplied length, so it's bounded by what's actually been allocated. No additional vector from this PR's scope.
- Dedup check at `eval.rs:15403-15411` uses `HashSet<String>` over the (now-bounded) keys. Was a secondary DoS before this PR (1B-entry array of `'x'` would hash 1B times); now bounded to 1M. Good.

## Verdict

The fix is correct and targeted, the cap value is well chosen, and TypeError is the right error type. Posting as **comment** (not requesting changes) because the engineering bar is met for the immediate DoS mitigation. The major findings are about robustness/consistency/test coverage, not about the patch being wrong.

Strongly recommend, before merging:
1. Close PR #116 as duplicate (or close this one — no preference).
2. Add the `test262-extra` regression tests called out above.
3. Fold the silent-0 Infinity/negative branch into the TypeError throw, OR document explicitly why these are intentionally silent.
4. Hoist the constant to module scope.

Issue #67, opened 2026-04-16
Bytecode compiler to eliminate AST-walking overhead


pmatos:
## Description

The tree-walking interpreter re-traverses AST nodes on every execution, paying match-dispatch, `Completion` enum wrapping, and recursive `eval_expr` call overhead on every single expression. A bytecode compiler that lowers the AST to a flat instruction sequence would eliminate this overhead and enable further optimizations (register allocation, peephole opts, superinstructions).

## Motivation

Profiling the mandreel benchmark (#54) shows that even after adding fast paths for Number arithmetic, typed array access, and identifier resolution, the per-operation cost is dominated by:
- `eval_expr` match dispatch (~30 variants)
- `Completion` enum construction/destruction on every operation
- Recursive function call overhead for nested expressions
- Branch prediction misses from polymorphic dispatch

An empty `for` loop of 1M iterations takes ~1,250ms (1.25μs/iter), which is almost entirely dispatch overhead since the loop body is empty.

## Scope

- Design a bytecode instruction set (stack-based or register-based)
- Implement a compiler pass: AST → bytecode
- Implement a bytecode interpreter (dispatch loop)
- Maintain spec compliance (test262 100%)

## References

- #54 (mandreel benchmark timeout — parent issue)
- PR #65 (interpreter fast paths — partial mitigation)

pmatos:
Implemented the next conservative vertical slice in PR #349. I extended the existing opt-in stack VM instead of attempting a register-VM rewrite or enabling bytecode by default: function-scoped `var`, captured identifier assignment/update, and basic numeric `for`/`while` loops now compile, while lexical loops, abrupt loop control, and unsupported syntax retain whole-body fallback. This keeps the compliance boundary explicit while delivering measured improvements of 2.35x for an empty 1M-iteration loop and 1.48x for a summing loop.

Validation was clean across Rust quality gates, 320 unit tests, the smoke oracle, 8 custom tests, and 3,021 targeted bytecode test262 scenarios. The full default run reported 22 existing Intl locale-data failures that reproduce with bytecode disabled. The first full bytecode run had those same 22 plus 12 Array.lastIndexOf timeout artifacts during shared-host load; every timed-out case passed in isolated bytecode reruns.

Issue #54, opened 2026-04-14
JetStream 3: mandreel benchmark times out (>120s)


pmatos:
## Description

The JetStream 3 `mandreel` benchmark (Octane suite) times out at 120s with a single iteration. This is the Mandreel C-to-JS compiler output — an extremely large, computation-heavy workload that stress-tests raw execution throughput.

## Reproducing

```bash
# Clone JetStream 3
gh repo clone WebKit/JetStream /tmp/JetStream -- --depth 1

# Run the benchmark (will time out)
cargo build --release
uv run python scripts/run-jetstream.py --test mandreel --iterations 1 --timeout 120
```

Or manually:
```bash
cat /tmp/JetStream/Octane/mandreel.js > /tmp/bench.js
cat >> /tmp/bench.js << 'EOF'
const start = Date.now();
const benchmark = new Benchmark();
benchmark.runIteration(0);
console.log("time:", Date.now() - start, "ms");
EOF
timeout 120 ./target/release/jsse /tmp/bench.js
```

The file is ~5MB of dense JavaScript with heavy arithmetic loops, array access, and bitwise operations — the kind of code that benefits most from JIT compilation. Even so, finishing a single iteration in <120s would be a useful target.

## Affected JetStream 3 benchmark

- mandreel (Octane, 80 default iterations)

pmatos:
## Direct perf attempt — measured, still blocked on the JIT/bytecode epic

Measured a single `runIteration(0)` on a 61-core box (contention-free):

| Binary | mandreel time |
|--------|---------------|
| `origin/main` (c7ad21d) | **>260s** (timed out at 260s) |
| combined-perf (this session's #71+#72+#73+#74+#125 merged) | **>220s** (timed out at 220s) |

**Conclusion:** the recent interpreter micro-optimizations (prototype-data inline cache #71, hoist-analysis cache #72, `has_simple_params` cache #73, property-key interning #74, array fast paths #125) do **not** meaningfully move mandreel. It's ~5MB of dense arithmetic/bitwise/array C-to-JS output whose cost is raw per-operation **dispatch throughput** — the tree-walker pays match-dispatch + `Completion` wrapping on every operation. Getting a single iteration under 120s needs the **bytecode VM / JIT** (#67), not property/call/array micro-opts.

Marking **ready-for-human**: blocked on the bytecode-compiler epic (#67). Reproduction (with a `performance` shim the Octane harness needs) is in `/tmp/bench/mandreel.js`.

[...]

PR #130, opened 2026-05-04
[bytecode] Foundation: stack VM scaffold behind --bytecode (refs #67)

pmatos:
## Summary

Foundation for the bytecode compiler tracked in #67. This PR delivers Phase 1 + an initial slice of Phase 2 from the implementation plan: a stack-based VM, an eligibility membrane, and the scaffolding to dispatch eligible function bodies through compiled chunks. Behind the new `--bytecode` flag, off by default — zero behaviour change without the flag.

**Important framing: this PR does not yet deliver the perf win that motivates #67.** The remaining work — control flow (`if`/`while`/`for`), function calls, member access, slot-based locals — is queued for follow-up PRs per the plan in `.ultraplan/bytecode-compiler-dispatch.md`. The point of landing this now is to get architectural review on the eligibility membrane, the cache placement on `JsObjectData`, the `eval_binary`/`eval_unary` reuse pattern, and the VM dispatch shape before more code is built on potentially-wrong choices.

## What's in this PR (4 commits, ~750 net lines)

1. **Phase 0 + Phase 1 scaffold** (f4a3bdc) — `src/interpreter/bytecode/` module, `Op` enum, `Chunk` + `Constant` types, stack VM dispatch loop, witness counter (`bytecode_chunks_executed`), `BytecodeCacheState` on `JsObjectData` (`Untried | Ineligible | Compiled(Rc<Chunk>)`), dispatch wired into `call_function_inner`'s sync-User path via new `Interpreter::dispatch_body`. `--bytecode` CLI flag plumbed through to `Interpreter` and the test runner. Phase 1 opcodes: `LoadConst`, `LoadUndefined`, `Return`, `ReturnUndefined`, `Add`.
2. **Phase 2 expression coverage** (37c76b0) — full arithmetic (`Sub`/`Mul`/`Div`/`Mod`/`Pow`), comparison/equality (`Eq`/`NotEq`/`StrictEq`/`StrictNotEq`/`Lt`/`Gt`/`LtEq`/`GtEq`), bitwise (`BitAnd`/`BitOr`/`BitXor`/`Shl`/`Shr`/`UShr`), unary (`Neg`/`Plus`/`Not`/`BitNot`), `Void`, `Conditional` (`?:`), logical short-circuit (`&&`/`||`/`??`) via `JumpIfTruthyKeep`/`JumpIfFalsyKeep`/`JumpIfNotNullishKeep`. `Pop`, `LoadTrue`/`LoadFalse`/`LoadNull`. Forward jumps with `i16` operand and patch machinery.
3. **`LoadName` for identifier reads** (1a66ef0) — separate `names: Vec<Rc<str>>` pool on `Chunk` keeps identifier names distinct from string-literal constants (so a constant `"x"` and the identifier `x` can't accidentally alias). Threads the runtime `env` through to `Interpreter::resolve_identifier` (made `pub(super)`), preserving strict-mode `ReferenceError` behaviour for undeclared names.
4. **`StoreName` + `ExpressionStatement` + `Sequence`** (ca803b8) — simple identifier assignment via `Interpreter::resolve_identifier_ref` + `put_value_by_ref` (both made `pub(super)`); strict / sloppy / with-scope / TDZ / const-assign all behave identically to the AST path because the bytecode delegates straight to the existing helpers. `Statement::Expression` (compile + `Pop`), `Statement::Empty` (no-op), `Expression::Sequence` / `Expression::Comma` (compile each, `Pop` between, last value remains — the parser wraps `(x = 99)` as a single-element `Sequence`, so this is needed for nearly every parenthesised expression).

## Design decisions worth review

- **Eligibility = compile_body returns `Ok`.** No separate eligibility predicate. The compiler is conservative; anything it doesn't recognise (any non-`Return` / non-`Expression` / non-`Empty` statement, any unsupported expression node) returns `CompileError::Unsupported(...)` and the function falls back to AST forever via `BytecodeCacheState::Ineligible`.
- **Cache lives on `JsObjectData`, not `JsFunction::User`.** `JsObjectData` is wrapped in `Rc<RefCell<>>` in the slab so a plain `Option<Rc<Chunk>>` field works without inner `RefCell`. Avoids touching the ~40 `JsFunction::User { ... }` construction sites in `eval/literals.rs`.
- **Opcode handlers reuse `eval_binary` / `eval_unary` / `resolve_identifier` / `resolve_identifier_ref` / `put_value_by_ref`.** Spec semantics — string concat in `+`, ToPrimitive, BigInt, NaN comparisons, ToInt32 / ToUint32, ToBoolean, with-scope handling, TDZ, const-assign-throws, sloppy-mode global creation — match the AST path bit-for-bit because they share the same helper functions. The visibility of those helpers is bumped to `pub(super)`; no behaviour change.
- **Witness counter on `Interpreter`.** `bytecode_chunks_executed` increments in `run_chunk`. Tests assert which path was taken — so `end_to_end_ineligible_function_falls_back_to_ast` actually verifies the fallback, and `end_to_end_literal_return_takes_bytecode_path` actually verifies the bytecode runs (instead of silently passing because both modes return the same value).
- **Generators / async / `arguments` / `with` / direct `eval` / `try` are blocked at the eligibility layer.** Plan calls these out as Phase 4+ concerns. Compile_body refuses every non-trivial statement today, so nothing reaches them through bytecode.

## Test plan

- [x] `cargo test --bin jsse` — 99/99 pass, including 4 unit + 25 integration tests inside the new `bytecode` module. Mix of (a) hand-built `Chunk` execution to lock down the VM contract, (b) compiler round-trip through `compile_body`, and (c) end-to-end `var __r = (function(){...})();` integration tests with witness-counter assertions on which path executed.
- [x] `cargo build --release` — clean.
- [x] test262 expression-heavy subset under `--bytecode`: `Math` + `language/expressions` + `language/statements/return` = **21,849 / 21,849 (100%)**, zero regressions vs `origin/main:test262-pass.txt` baseline.
- [ ] Full `--bytecode` test262 — running now, log will be posted as a comment when complete (running locally; result expected to be 100% modulo the 32 pre-existing regressions in `RegExp/prototype/Symbol.match*` that exist on `origin/main` independent of this branch — verified by checkout-and-test on commit `9d2a268`).
- [x] test262 baseline (no `--bytecode`) on the same subset — identical 100% result, confirming zero behaviour change with the flag off.

## What's intentionally NOT in this PR

- Statement support beyond `Return` / `Expression` / `Empty` (so no `if`/`while`/`for`/`try`/`var`/`let`/`const`)
- `Call`, `New`, member access, `Update` (`++`/`--`), compound assign (`+=`)
- `typeof` operator, `delete`, `super`, `NewTarget`, `this`
- Slot-based locals — every `LoadName` still walks the env chain via `resolve_identifier`
- Top-level script bodies — only function bodies are compiled
- Differential test262 CI workflow

These are the next vertical slices in `.ultraplan/bytecode-compiler-dispatch.md` and will land as separate PRs. The empty-1M-loop benchmark cited in #67 will not move until at least Phase 5.1 (slot-based locals) lands; that's a deliberate deferral so this PR stays small and reviewable.

Refs #67. Refs #54 (mandreel timeout — perf gap that the eventual full pipeline targets).


pmatos:
## Full `--bytecode` test262 validation

```
files:       51,525
scenarios:   99,020
pass:        98,985 (99.96%)
regressions: 35
new_passes:  0
```

**Net result: zero new regressions from this branch.** Of the 35 reported regressions:

- **32 are pre-existing on `origin/main` HEAD** — all in `RegExp/prototype/Symbol.match*` and `Symbol.matchAll/*`. Verified during Phase 1 development by reverting to commit `9d2a268` (origin/main HEAD) and re-running the same tests in plain AST mode; the same 32 fail. Likely traceable to PR #124 (`e5945e3 [perf] Intern Environment.global_object and WithObject._object`). These should be tracked separately from this PR.

- **3 are load-flaky `Atomics.waitAsync` tests** that pass when re-run in isolation:
  - `built-ins/Atomics/waitAsync/bigint/true-for-timeout.js` (default + strict)
  - `built-ins/Atomics/waitAsync/true-for-timeout.js:strict`
  
  Re-run with `-j 4` on just those 4 scenarios → 4/4 pass. Consistent with the existing "Flaky test under load" pattern noted in the project's memory file. The waitAsync timing is sensitive to system load; the full-suite run was sharing the box with a JetStream `mandreel` benchmark in another worktree.

So zero **bytecode-specific** regressions. Both `--bytecode` and AST modes produce identical results on every test that passes baseline.

Issue #181, opened 2026-06-19
[test262] Source-phase import re-export semantics unimplemented (import source / <module source>)


pmatos:
## Summary

The **source-phase imports** proposal (`import source x from "…"`) is not implemented at runtime in jsse. Three new tests in the bumped test262 snapshot (`de8e621c`) exercise its **re-export** semantics and all fail with the identical error `Cannot resolve bare module specifier '<module source>'` (from `resolve_module_specifier`, `src/interpreter/mod.rs:2155`).

**Impact:** 3 test files / **3 scenarios** (module mode), features `source-phase-imports` + `source-phase-imports-module-source`.

> Historical context: this was partially tracked by the now-**closed** #91 ("implement import defer and source phase imports proposals"). Import-defer landed; source-phase imports were only **stubbed** (parser + a throwing `.AbstractModuleSource`). This issue tracks the remaining runtime work specifically surfaced by these re-export tests.

## Current state in jsse (stubs only)

- Parser accepts `import source X from "…"` → `ImportSpecifier::SourcePhase` (`src/parser/modules.rs:51–107`).
- `.AbstractModuleSource` constructor always throws TypeError (`src/interpreter/mod.rs:686–744`).
- Static linker **rejects** the binding: `ImportSpecifier::SourcePhase(_) => Err(TypeError "Source phase imports are not supported…")` (`mod.rs:1927–1931`).
- Dynamic `import.source()` always rejects with SyntaxError (`src/interpreter/eval.rs:1190–1196`).
- No `[[ModuleSource]]` slot; no `~source~` branch in `resolve_export`/`resolve_export_binding`; no source branch in module-namespace `[[Get]]`.

## Why all three fail the same way (one cause)

Each test imports a fixture containing `import source x from '<module source>'; export { x };`. `<module source>` is a test262 host convention (feature `source-phase-imports-module-source`) for a host-provided module whose `[[ModuleSource]]` is an `%AbstractModuleSource%`. jsse can't resolve that host specifier, and even a *resolvable* source-phase target hits the `SourcePhase` TypeError stub (verified directly). So none of the re-export-specific spec steps the tests target are ever reached:

- `language/module-code/ambiguous-export-bindings/namespace-unambiguous-if-import-source-and-export.js`
- `language/module-code/source-phase-import/reexport-source-binding-named-import.js`
- `language/module-code/source-phase-import/reexport-source-binding-namespace-get.js`

[...]

Issue #180, opened 2026-06-19
[test262][intl] Implement ECMA-402 legacy-constructor normative-optional behavior (%Intl%.[[FallbackSymbol]], Chain/Unwrap for NumberFormat & DateTimeFormat)


pmatos:
## Summary

jsse implements **none** of the ECMA-402 normative-optional **"legacy constructor"** feature. A grep across `src/` for `FallbackSymbol`, `IntlLegacyConstructedSymbol`, `ChainNumberFormat`/`ChainDateTimeFormat`, `UnwrapNumberFormat`/`UnwrapDateTimeFormat` returns nothing.

**Impact:** 8 test files / **16 scenarios** (default + strict), feature `intl-normative-optional`.

## What the spec requires

1. The `%Intl%` object has a **per-realm `[[FallbackSymbol]]`** internal slot — a fresh `Symbol` with `[[Description]]` `"IntlLegacyConstructedSymbol"`.
2. `Intl.NumberFormat` / `Intl.DateTimeFormat`, when called **without `new`** (newTarget undefined) on a receiver that is an instance of the corresponding prototype, run `ChainNumberFormat`/`ChainDateTimeFormat`: return `this` (not a fresh object) and define an own property keyed by `[[FallbackSymbol]]` holding the freshly-initialized formatter, descriptor `{writable:false, enumerable:false, configurable:false}`.
3. The prototype methods (`resolvedOptions`, `format`, …) run `UnwrapNumberFormat`/`UnwrapDateTimeFormat`: if the receiver lacks the internal slot but has the `[[FallbackSymbol]]` own property, use that property's value as the formatter.

## What jsse does instead

- **Constructor closures ignore `this`/`new_target`.** `numberformat.rs` (~L4419–4452) and `datetimeformat.rs` (~L5710–5751) always `create_object_id()` and return a brand-new object regardless of call-vs-construct. The discriminator (`self.new_target`) is already maintained correctly (`eval.rs` ~L12366) but unused here.
- **No `[[FallbackSymbol]]` slot.** `Realm` (`types.rs:311`) has the intl prototype fields but no per-realm legacy symbol.
- **No Unwrap step.** `extract_dtf_data` (`datetimeformat.rs:4172`) and NumberFormat `resolvedOptions` (`numberformat.rs:3559`) only accept a `this` carrying its own `IntlData` slot.

## Symptom → cause mapping (all one feature gap)

[...]

Issue #179, opened 2026-06-19
[test262] Implement Error.prototype.stack get/set accessor (error-stack-accessor); remove own "stack" data property on instances


pmatos:
## Summary

jsse does **not** implement the normative-optional **`error-stack-accessor`** feature (the merged "Error Stacks" proposal). There is no `get`/`set` accessor pair for `stack` on `%Error.prototype%`. Instead, jsse installs an **own empty-string `stack` data property** on every Error-family instance at construction time. The new test262 suite `built-ins/Error/prototype/stack/` (added in the upstream snapshot we just bumped to, `de8e621c`) exercises the accessor and fails across the board.

**Impact:** 35 test files / **70 scenarios** (default + strict), 0 currently passing. Not regressions — these tests are new in the bumped snapshot.

## Root cause (one cause, two observable symptoms)

1. **Accessor absent** (33 files). Every test begins with `Object.getOwnPropertyDescriptor(Error.prototype, 'stack')`, which is `undefined` in jsse, so getter/setter probes throw `TypeError: Cannot read properties of undefined (reading 'get'|'set')`, and `prop-desc.js` fails "getter should be a function".
2. **Wrong own data property** (`instance-no-own-stack.js`, `instance-not-enumerable.js`). Because jsse inserts an own `stack` data property, `hasOwnProperty(err,'stack')` is `true` (must be `false`) and `getOwnPropertyNames(err)` includes `"stack"` (must not).

Both vanish once the accessor is installed on `Error.prototype` and the per-instance inserts are removed.

## Required shape (per the tests' `info:` blocks)

[...]

Issue #169, opened 2026-06-11
TypedArray.prototype.slice throws on immutable ArrayBuffer when species target shares the buffer


pmatos:
## Summary

When `%TypedArray%.prototype.slice` creates its output via `Symbol.species` and the species returns a typed array backed by the **same immutable `ArrayBuffer`** (different byte offset), jsse throws `TypeError: Cannot modify an immutable ArrayBuffer`. The spec requires the element-by-element copy to succeed. Affects all TypedArray element types (verified Int8/Uint8/Int32/Float64).

## Failing test

- `test262/test/built-ins/TypedArray/prototype/slice/speciesctor-return-same-buffer-with-offset.js` (+`:strict`)

Not in the `origin/main` baseline — known-failing, not a regression. The non-immutable factory cases (`makePassthrough`/`makeArray`/…) already pass; only the `makeImmutableArrayBuffer` factory (which uses `ArrayBuffer.prototype.transferToImmutable`, from issue #140) trips it.

## Reproduction

```js
var TA = Int8Array;
var imm = new TA([10,20,30,40,50,60]).buffer.transferToImmutable();
var ta = new TA(imm);
ta.constructor = { [Symbol.species]: function(){ return new TA(ta.buffer, 2*TA.BYTES_PER_ELEMENT); } };
ta.slice(1, 4);
// jsse:   TypeError: Cannot modify an immutable ArrayBuffer
// expect: returns [20, 20, 20, 60]   (length 4: species view covers buffer[2..6])
```

The expected `[20, 20, 20, 60]` is the result of the overlapping copy: src=ta@offset1, target=species view@offset2 over the same buffer; the first 3 elements are overwritten by the forward byte copy and the trailing `60` is left untouched.

## Why it should not throw (spec)

Per `sec-%typedarray%.prototype.slice`, the same-type copy path is **internal** (the test cites it):

```
14.g.ix. Repeat, while targetByteIndex < endByteIndex,
  1. value = GetValueFromBuffer(srcBuffer, srcByteIndex, uint8, true, unordered)
  2. SetValueInBuffer(targetBuffer, targetByteIndex, uint8, value, true, unordered)
```

The immutable-ArrayBuffer proposal applies its mutability guard at the user-facing `[[Set]]` / `IntegerIndexedElementSet` level, **not** at the internal `SetValueInBuffer` used by `slice`/`set`/`sort`. jsse's guard fires too low (inside the slice byte-copy), so it rejects a copy the spec permits.

[...]

PR #184, opened 2026-06-20
[intl] PluralRules: thread [[CompactDisplay]] into select/selectRange; fix selectRange ignoring notation (#174)

### LINKED ISSUE #174: Intl.PluralRules: select/selectRange should pass [[CompactDisplay]] into PluralRuleSelect
Deferred from #171 (which reads & surfaces the `compactDisplay` option).

Original feedback by Codex on PR #171:

> Storing `compact_display` makes `new Intl.PluralRules(..., { notation: "compact", compactDisplay: "long" }).resolvedOptions()` report `long`, but the `select`/`selectRange` paths still ignore this slot when computing categories. ECMA-402's ResolvePlural/ResolvePluralRange pass `[[CompactDisplay]]` into `PluralRuleSelect`, so for locales where long and short compact forms have different plural behavior, callers now observe the option as accepted while selections are still computed without it.

**Spec:** ECMA-402 `PluralRuleSelect ( locale, type, notation, compactDisplay, s )` does take `compactDisplay`, and `Intl.PluralRules` §17.4 notes `[[CompactDisplay]]` "can in some cases influence plural form selection". So `select`/`selectRange` should thread `[[CompactDisplay]]` into category computation.

**Context:** `src/interpreter/builtins/intl/pluralrules.rs` — `select` (~line 230) and `selectRange` call `number_to_plural_operands_with_notation(n, &notation)`, which receives `notation` but not `compactDisplay`. Category selection is delegated to ICU4X (`PluralRules::category_for(operands)`), which operates on numeric operands and does not currently take a short/long compact-display input.

**Why deferred from #171:** scope-creep + no failing test. #171's scope is reading and surfacing the option (fixing 3 failing test262 option-reading tests); it does not regress `select` (which already ignored `compactDisplay`, since the slot did not exist before). No test262 test exercises `select`/`selectRange` with `compactDisplay`, and the full `intl402/PluralRules` suite passes 106/106 with #171. Implementing compactDisplay-aware selection is a distinct feature that needs:
- a concrete locale + value where short vs long compact display yields a different plural category (to drive a test, since `PluralRuleSelect` is implementation-defined), and
- a mechanism for ICU4X to honour compactDisplay in category selection (it may not expose one — the compact exponent already flows through the `c`/`e` operand via `notation: "compact"`).

Without such a case, a "fix" would be speculative. Filing for follow-up once a demonstrable selection difference (and ICU4X support) is identified.

pmatos:
Closes #174.

## Summary

Issue #174 asked that `Intl.PluralRules.prototype.select`/`selectRange` thread `[[CompactDisplay]]` into `PluralRuleSelect`. Investigation established two things:

1. **compactDisplay is a no-op for category selection** under CLDR/ICU4X. `icu_plurals` `PluralOperands` carries the CLDR operands `n,i,v,w,f,t,c,e`; the `c`/`e` operand is the *compact exponent* (identical for "1K" and "1 thousand"). There is no operand field and no `category_for` API distinguishing short vs long. ECMA-402 `PluralRuleSelect` is implementation-defined; the §17.4 "can in some cases influence" note is a hedge. Empirically, node reports **0** category differences for short vs long across many locales/values.

2. **`selectRange` had a genuine adjacent bug**: it ignored `notation` entirely, building operands via the plain `number_to_plural_operands` while `select` used the notation-aware builder. So compact/scientific/engineering *ranges* were categorized from the standard integer value instead of the compact form — a spec deviation (ResolvePluralRange computes endpoint categories with the construct's notation).

This PR does both (per the agreed "implement + fix bug" scope):

- Threads `compact_display` through the operand path in `select` and `selectRange` to mirror `PluralRuleSelect(locale, type, notation, compactDisplay, s)`. It is documented as a no-op for selection under current locale data (the param is intentionally unconsumed by the operand model).
- Fixes `selectRange` to honor `notation` (and `compactDisplay`) exactly as `select` does.

## Change
`src/interpreter/builtins/intl/pluralrules.rs` (+33/-4):
- `number_to_plural_operands_with_notation` gains a documented `compact_display` parameter.
- `select` threads `compact_display`.
- `selectRange` now destructures + honors `notation`/`compact_display` (the bug fix).

## Tests (test262-extra)
- **`Intl-PluralRules-selectRange-notation.js`** — oracle-free regression test: the invariant `selectRange(v,v) === select(v)` for compact-discriminating `fr` values (e.g. 1_100_000, where compact -> "many" but standard -> "other"). This **fails on the pre-fix code** (selectRange returned "other") and **passes after**. A guard asserts the chosen values stay notation-discriminating so CLDR data drift fails loudly rather than silently.
- **`Intl-PluralRules-compactDisplay-noop.js`** — characterization: `compactDisplay: "short"` vs `"long"` yields identical categories for `select` and `selectRange` across locales/values, and `resolvedOptions()` still surfaces the option. Green before and after the threading, locking in that threading changes nothing observable.

## Validation
- `./scripts/lint.sh` (rustfmt + clippy `-D warnings`): clean.
- `intl402/PluralRules` on the de8e621c base: **110/110, 0 regressions, 0 failures** (incl. the 2 new test262-extra tests).
- Full test262 suite: **0 regressions** (run in progress on the PR base; final number in a follow-up comment).

### DIFF
--- a/src/interpreter/builtins/intl/pluralrules.rs
+++ b/src/interpreter/builtins/intl/pluralrules.rs
@@ -30,7 +30,21 @@ fn number_to_plural_operands(n: f64) -> PluralOperands {
     PluralOperands::from(&decimal)
 }
 
-fn number_to_plural_operands_with_notation(n: f64, notation: &str) -> PluralOperands {
+/// Builds CLDR plural operands for `n` under the given `notation`.
+///
+/// `compact_display` mirrors ECMA-402 `PluralRuleSelect(locale, type, notation,
+/// compactDisplay, s)`: it is threaded through so `select`/`selectRange` match
+/// the spec's selection signature. Under CLDR/ICU4X the plural operands
+/// (including the compact exponent `c`) are identical for `"short"` and `"long"`
+/// compact forms ("1K" and "1 thousand" share the same operands), so the
+/// parameter does not affect operand construction — it is intentionally
+/// unconsumed here. (Verified: node and icu_plurals report identical categories
+/// for short vs long.)
+fn number_to_plural_operands_with_notation(
+    n: f64,
+    notation: &str,
+    _compact_display: Option<&str>,
+) -> PluralOperands {
     if n.is_nan() || n.is_infinite() {
         return PluralOperands::from(0u64);
     }
@@ -192,6 +206,7 @@ impl Interpreter {
                         locale,
                         plural_type,
                         notation,
+                        compact_display,
                         ..
                     }) = data
                     {
@@ -227,7 +242,11 @@ impl Interpreter {
                             }
                         };
 
-                        let operands = number_to_plural_operands_with_notation(n, &notation);
+                        let operands = number_to_plural_operands_with_notation(
+                            n,
+                            &notation,
+                            compact_display.as_deref(),
+                        );
                         let cat = rules.category_for(operands);
                         return Completion::Normal(JsValue::String(JsString::from_str(
                             plural_category_to_str(cat),
@@ -258,6 +277,8 @@ impl Interpreter {
                     if let Some(IntlData::PluralRules {
                         locale,
                         plural_type,
+                        notation,
+                        compact_display,
                         ..
                     }) = data
                     {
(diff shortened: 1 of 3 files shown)


pmatos:
Full test262 run on the PR base (test262 `de8e621c`):

**99,424 / 99,515 (99.91%), 91 failures** — identical to the documented `de8e621c` baseline state.

**No PluralRules test fails or regresses.** Focused `intl402/PluralRules` (+ the 2 new test262-extra tests) is **110/110, 0 failures**.

The runner reports 8 "regressions" vs `origin/main:test262-pass.txt`, but all 8 are `intl402/NumberFormat/intl-legacy-constructed-symbol*` — that is **#180** (Intl normative-optional legacy-ctor `[[FallbackSymbol]]`), a pre-existing artifact of the stale `c42f56da` baseline vs the strengthened `de8e621c` snapshot, **not introduced by this change** (which only touches `pluralrules.rs`). The remaining failures are the known `de8e621c` set (#179 / #180 / #181 / #169). 185 new passes are baseline lag from the submodule bump.

PR #252, opened 2026-07-16
Deepen builtin accessor registration with define_getter helper

pmatos:
## Refactoring goal

Surfaced by an architecture audit (mattpocock/skills `improve-codebase-architecture`). The audit's top low-risk / high-leverage **deepening opportunity**: the accessor-install *dance* is open-coded at every built-in getter site.

The engine already has `Interpreter::define_method`, which collapsed the method-install boilerplate (`create_function` → `insert_builtin`) into one deep call. The **accessor** half never got the same treatment: each read-only getter still hand-wrote

```rust
let x_getter = self.create_function(JsFunction::native("get x".to_string(), 0, |..| { .. }));
self.get_object_cell_expect(proto_id).borrow_mut().insert_property(
    "x".to_string(),
    PropertyDescriptor { value: None, writable: None, get: Some(x_getter),
                         set: None, enumerable: Some(false), configurable: Some(true) },
);
```

The **interface** (install a read-only getter) was buried under an **implementation** repeated ~dozens of times — a shallow, leaky seam. Deletion test: removing the six-field literal concentrates the descriptor shape in one module rather than scattering it.

## What changed

- **New `Interpreter::define_getter(target_id, name, f)`** (`src/interpreter/mod.rs`), companion to `define_method`. Mints the `get {name}` native (length 0, per spec) and installs `{ get, enumerable: false, configurable: true }` in one place. Returns the getter value.
- **Adopted at 22 sites across 5 core built-in modules:**
  - `typedarray.rs` — `ArrayBuffer` / `SharedArrayBuffer` / `%TypedArray%` / `DataView` getters (`byteLength`, `byteOffset`, `length`, `buffer`, `maxByteLength`, `resizable`, `growable`, `detached`, `immutable`)
  - `collections.rs` — `Map.prototype.size`, `Set.prototype.size`
  - `regexp.rs` — `RegExp.prototype.flags`, `RegExp.prototype.source`
  - `number.rs` — `Symbol.prototype.description`
  - `disposable.rs` — `DisposableStack` / `AsyncDisposableStack` `disposed`
- **Net ~280 fewer lines** (539 insertions, 821 deletions).

### Intentionally out of scope
- **Symbol-keyed getters** (`get [Symbol.species]`, `get [Symbol.toStringTag]`) and **get+set accessors** (e.g. `Iterator.prototype.constructor`) are left untouched — the helper is read-only and string-keyed by design.
- **Temporal / Intl getters** (~30) use a different registration shape (`PropertyDescriptor` field order differs, wrapped in scope blocks, several are loop/table-driven). Converting them cleanly is a separate mechanical pass; deferred to keep this diff uniform and fully verified. Good follow-up.

## Tests

Built test-first (TDD, mattpocock/skills `tdd`):

- **New unit test** `define_getter_installs_a_correctly_shaped_accessor` (`src/interpreter/tests.rs`), mirroring the existing `define_method_installs_a_correctly_shaped_builtin`. Pins the raw stored descriptor (`get` present, no `set`/`value`/`writable`, `enumerable: false`, `configurable: true`), the getter's `.name == "get answer"` and `.length == 0`, that the returned value is the installed getter, and that it computes on call. Written **RED** (helper absent → compile error) → **GREEN**.
- **Behavior preservation** — the converted getters' `.name` / `.length` / `enumerable` / `configurable` are asserted directly by test262:
  - Full **`built-ins/` sweep: 47,031 scenarios, 100%, 0 regressions**
  - Per-domain: TypedArray, TypedArrayConstructors, ArrayBuffer, SharedArrayBuffer, DataView, Map, Set, WeakMap, WeakSet, Number, Symbol, RegExp, DisposableStack, AsyncDisposableStack — all 100% / 0 regressions
- `cargo test --release` — 291 passed (+ test262 smoke oracle)
- `./scripts/lint.sh` — rustfmt + `clippy -D warnings` clean
- `run-custom-tests.py` — 6/6

### DIFF
--- a/src/interpreter/builtins/collections.rs
+++ b/src/interpreter/builtins/collections.rs
@@ -358,43 +358,26 @@ impl Interpreter {
         );
 
         // Map.prototype.size (getter)
-        let size_getter = self.create_function(JsFunction::native(
-            "get size".to_string(),
-            0,
-            |interp, this, _args| {
-                if let JsValue::Object(o) = this
-                    && let Some(obj) = interp.get_object_cell(o.id)
-                {
-                    let borrowed = obj.borrow();
-                    let is_map = borrowed.map_data().is_some() && borrowed.class_name == "Map";
-                    let map_data = if is_map {
-                        borrowed.map_data().cloned()
-                    } else {
-                        None
-                    };
-                    drop(borrowed);
-                    if let Some(entries) = map_data {
-                        let count = entries.iter().filter(|e| e.is_some()).count();
-                        return Completion::Normal(JsValue::Number(count as f64));
-                    }
+        self.define_getter(proto_id, "size", |interp, this, _args| {
+            if let JsValue::Object(o) = this
+                && let Some(obj) = interp.get_object_cell(o.id)
+            {
+                let borrowed = obj.borrow();
+                let is_map = borrowed.map_data().is_some() && borrowed.class_name == "Map";
+                let map_data = if is_map {
+                    borrowed.map_data().cloned()
+                } else {
+                    None
+                };
+                drop(borrowed);
+                if let Some(entries) = map_data {
+                    let count = entries.iter().filter(|e| e.is_some()).count();
+                    return Completion::Normal(JsValue::Number(count as f64));
                 }
-                let err = interp.create_type_error("Map.prototype.size requires a Map");
-                Completion::Throw(err)
-            },
-        ));
-        self.get_object_cell_expect(proto_id)
-            .borrow_mut()
-            .insert_property(
-                "size".to_string(),
-                PropertyDescriptor {
-                    value: None,
-                    writable: None,
-                    get: Some(size_getter),
-                    set: None,
-                    enumerable: Some(false),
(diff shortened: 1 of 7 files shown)


pmatos:
### Note on `new_passes: 176` in the built-ins sweep

This is **pre-existing baseline drift, not a behavior change** from this PR (the runner compares against `origin/main:test262-pass.txt`):

- **174 of 176** are in directories this PR never touches (Temporal/Intl/etc.). Untouched code cannot be affected — the per-domain runs showed only TypedArray had any deltas (2).
- The conversion is **name/length/attribute-identical by construction**: the mechanical converter only rewrote a block when the getter function name was exactly `"get " + propertyKey`, and `define_getter` reproduces precisely that via `format!("get {name}")` + the same read-only descriptor. So even the 2 TypedArray entries cannot originate from this diff.

Net behavior change: **0**. `regressions: 0` across all 47,031 built-ins scenarios.

README pass count is intentionally left unchanged (pass rate is unchanged; only `built-ins/` was run here, not the full suite).

PR #257, opened 2026-07-16
ci: run Miri over unsafe-adjacent modules

### LINKED ISSUE #194: Tooling: run Miri over the unsafe-adjacent modules
Follow-up from #192. The engine has ~29 `unsafe` blocks (`src/interpreter/types.rs`, `src/interpreter/builtins/atomics.rs`); Miri detects UB (aliasing, uninit reads, OOB) that clippy/semgrep cannot.

Deferred from the CI-hardening PR because it could not be validated locally (miri not installed; the whole icu/chrono-heavy crate must compile as MIR regardless of the test filter) and, as a `schedule`/`workflow_dispatch` job, it would not run on the PR to self-validate.

Ready-to-use recipe:
```yaml
name: Miri
on:
  schedule: [{ cron: "0 5 * * 1" }]
  workflow_dispatch:
permissions:
  contents: read
jobs:
  miri:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@<sha> # v7
        with: { persist-credentials: false }
      - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # master
        with: { toolchain: nightly, components: miri }
      - run: cargo +nightly miri test property_map object_arena gc key_intern
        env: { MIRIFLAGS: -Zmiri-disable-isolation }
```

- [ ] Confirm the crate builds under Miri (icu/chrono FFI is only *linked*, not *called*, by these pure-Rust arena/GC/map tests — Miri fails on FFI execution, not presence).
- [ ] Tune the test filter so it exercises the `unsafe`-touching code and stays green.
- [ ] Keep it nightly + non-gating.

pmatos:
## Summary

- add a weekly and manually dispatched nightly Miri workflow
- run narrowly scoped arena, GC, property-map, key-interning, shared-buffer, and Atomics tests
- add direct unit coverage for shared-buffer byte casts and every supported atomic integer width
- ignore intentional Rc-cycle leaks while retaining Miri checks for aliasing, invalid access, and uninitialized data

## Validation

- complete workflow-equivalent Miri loop: green
- cargo fmt --check: green
- cargo clippy: green
- cargo build --release: green
- cargo test --release: 293 unit tests and smoke oracle green
- actionlint and zizmor: green
- targeted Atomics and SharedArrayBuffer test262: 986/986 green
- full test262: 99,546/99,568; 22 unrelated Intl regressions and 323 new passes relative to the current baseline. A representative regression was reproduced with a separately built, unmodified origin/main binary, confirming it is not caused by this cfg(test)-only Rust coverage.

Closes #194

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -1,0 +1,40 @@
+name: Miri
+
+on:
+  schedule:
+    - cron: "0 5 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  miri:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      MIRIFLAGS: -Zmiri-disable-isolation -Zmiri-ignore-leaks
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          persist-credentials: false
+
+      - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # master
+        with:
+          toolchain: nightly
+          components: miri
+
+      - name: Run Miri over unsafe-adjacent tests
+        run: |
+          set -euo pipefail
+          for test_filter in \
+            'interpreter::property_map::tests::' \
+            'interpreter::object_arena::tests::' \
+            'interpreter::gc::tests::' \
+            'interpreter::key_intern::tests::' \
+            'interpreter::types::shared_buffer_tests::' \
+            'interpreter::builtins::atomics::tests::' \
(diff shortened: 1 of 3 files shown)


pmatos:
Implementation note on the Miri filter choices: Cargo accepts only one positional test filter, so the issue recipe as written fails on object_arena as an unexpected argument. The workflow runs each filter separately and uses full module paths because bare gc also selects unrelated integration tests. Miri leak checking is disabled because the focused GC test intentionally constructs a self-referential Rc environment; the alternatives were to skip that useful GC test or change unrelated GC-test semantics. Direct shared-buffer and atomic-helper tests were added so the job actually executes the unsafe casts/pointers across all supported integer widths rather than merely compiling them.

PR #285, opened 2026-07-16
fix: preserve tagged template substitutions across GC

### LINKED ISSUE #274: AJV corpus exposes generated-validator result mismatches under sustained codegen
## Summary

The AJV v8.17.1 JSON-Schema-Test-Suite integration from #234 deterministically reports 12 incorrect validation results after the corpus has generated and executed many validators. Node passes the identical 5,480-test bundle.

## Reproduction

```sh
./scripts/run-library-tests.sh ajv --clean
```

JSSE summary after the #234 literal-rooting and unary-tail-position fixes:

```text
PASS: 5466  FAIL: 14  TOTAL: 5480
```

Two of those 14 are separate call-stack failures tracked independently. The remaining 12 are assertion mismatches:

- draft 7: remote fragment valid; escaped dependency invalid object 3
- draft 2019-09: 8 cases in \"dynamic evalation inside nested refs\"; oneOf-with-required both-invalid
- draft 2020-12: escaped JSON-pointer percent reference valid

Error split is 9 false negatives and 3 false positives. The failing option variants include both default optimized output and ES5/lines/optimize=0 output.

The exact same failures recur in a second fresh JSSE process using the cached identical bundle. A standalone AJV2019 reproducer for the oneOf schema passes on both JSSE and Node, so reduction needs to preserve the sustained corpus setup/order (likely generated-closure lifetime, GC pressure, or another cross-validator state interaction).

## Expected

All 5,480 registered AJV cases pass, matching Node.

pmatos:
## Summary

- root the tag, receiver, template object, and accumulated substitution values while later tagged-template substitutions execute arbitrary JavaScript
- add a forced-GC regression for an earlier substitution surviving evaluation of a later substitution
- restore AJV generated-validator correctness under sustained code generation

## Validation

- `cargo fmt --check`
- `cargo clippy`
- `cargo build --release`
- `cargo test --release` (293 unit tests + smoke oracle)
- `uv run python scripts/run-custom-tests.py` (6/6)
- tagged-template test262 + regression (50/50)
- AJV v8.17.1 corpus on a clean #280 composite (5,480/5,480; issue baseline 5,466/5,480)
- full test262 (99,546/99,568): the 22 failures are existing Intl locale-data failures reproduced by an exact `origin/main` binary; 323 unrelated new passes, and no issue-related regression

Closes #274

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -1308,24 +1308,41 @@ impl Interpreter {
                     }
                 };
 
+                // The tag, its receiver, and each substitution value are held
+                // in Rust locals until EvaluateCall invokes the tag. Later
+                // substitutions can run arbitrary JavaScript (and therefore a
+                // GC safepoint), so keep all previously evaluated values live.
+                let gc_frame = self.gc_root_frame();
+                self.gc_root_value(&func_val);
+                self.gc_root_value(&this_val);
                 let template_obj = self.get_template_object(tmpl);
+                self.gc_root_value(&template_obj);
 
                 let mut call_args = vec![template_obj];
                 for sub_expr in &tmpl.expressions {
                     match self.eval_expr(sub_expr, env) {
-                        Completion::Normal(v) => call_args.push(v),
-                        other => return other,
+                        Completion::Normal(v) => {
+                            self.gc_root_value(&v);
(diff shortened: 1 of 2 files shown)


pmatos:
@codex review

PR #324, opened 2026-07-19
fix: honor exotic Set in prototype-ignoring setters

### LINKED ISSUE #185: Honor exotic [[Set]] in SetterThatIgnoresPrototypeProperties (Error.stack, Iterator.constructor, Iterator[Symbol.toStringTag])
Deferred from #183: the `SetterThatIgnoresPrototypeProperties` step 5a (`Set(this, p, v, true)`) ignores the boolean returned by `set_property_value` and does not honor exotic `[[Set]]` semantics for a receiver that reports a writable own data descriptor but whose `[[Set]]` returns `false`.

Original feedback by @chatgpt-codex-connector on PR #183 (https://github.com/pmatos/jsse/pull/183):

> **Route existing stack writes through [[Set]]**
>
> When the receiver already has an own `stack` descriptor, `SetterThatIgnoresPrototypeProperties` requires `Set(thisValue, "stack", v, true)`. This branch only checks descriptor bits and then ignores the boolean from `set_property_value`; for a module namespace object that exports `stack`, `[[GetOwnProperty]]` reports a writable own data descriptor but its `[[Set]]` returns false, so `Object.getOwnPropertyDescriptor(Error.prototype, "stack").set.call(ns, "x")` incorrectly succeeds instead of throwing. Calling the existing `proxy_set`/Set path and throwing on false would preserve the exotic `[[Set]]` behavior.

## Context

- `src/interpreter/builtins/mod.rs` — `Error.prototype.stack` setter, the ordinary own-data-descriptor branch: it calls `cell.borrow_mut().set_property_value("stack", v)` and discards the boolean.
- `src/interpreter/types.rs` — `set_property_value` models exotic `[[Set]] = false` **only** for typed-array out-of-bounds indices and non-writable `Array.length`. For a module namespace (whose `[[Set]]` always returns `false`, §10.4.6.9) it is a non-mutating no-op that returns `true`, so the setter returns normally instead of throwing.
- The **same gap already ships** for the sibling `SetterThatIgnoresPrototypeProperties` setters in `src/interpreter/builtins/iterators.rs`: `Iterator.prototype.constructor` and `Iterator.prototype[Symbol.toStringTag]` use the identical `contains_key` + `set_property_value` shortcut.

## Why deferred

diminishing-returns — the trigger (`Object.getOwnPropertyDescriptor(Error.prototype,'stack').set.call(ns,'x')` on a module namespace that exports a binding named `stack`) is highly contrived, has **zero test262 coverage** (the 35 `built-ins/Error/prototype/stack/` files and the Iterator `weird-setter.js` files all use ordinary objects / Proxies), and fixing only the Error branch would be incomplete-by-design.

## Suggested fix

Factor a shared `Set(O, P, V, Throw=true) -> bool` path (the dispatch is currently inlined in `Reflect.set` and `Object.assign`, including the "module namespace exotic → false" branch) and route all three `SetterThatIgnoresPrototypeProperties` setters (Error.prototype.stack, Iterator.prototype.constructor, Iterator.prototype[Symbol.toStringTag]) through it, throwing on `false`.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 and Codex (codex-cli 0.130.0), which both returned DEFER / diminishing-returns._

pmatos:
## Summary

- reject module namespace writes through the canonical `[[Set]]` entry point
- route `Error.prototype.stack`, `Iterator.prototype.constructor`, and `Iterator.prototype[Symbol.toStringTag]` existing-property writes through that entry point
- add a module-namespace regression covering all three setters

## Testing

- `cargo fmt --check`
- `cargo clippy`
- `./scripts/lint.sh`
- `cargo build --release`
- `cargo test --release`
- `uv run python scripts/run-custom-tests.py`
- focused custom regression: 1/1
- focused Error/Iterator test262: 78/78
- `TZ=UTC uv run python scripts/run-test262.py`: 99,568/99,568

Closes #185

### DIFF
--- a/src/interpreter/builtins/iterators.rs
+++ b/src/interpreter/builtins/iterators.rs
@@ -651,11 +651,15 @@ impl Interpreter {
                                 PropertyDescriptor::data(v, true, true, true),
                             );
                         }
-                    } else if let Some(od) = interp.get_object_cell(this_id)
-                        && !od.borrow_mut().set_property_value(&prop_key, v)
-                    {
-                        let err = interp.create_type_error("Cannot set property");
-                        return Completion::Throw(err);
+                    } else {
+                        match interp.set_object_property(this_id, &prop_key, v, this) {
+                            Ok(true) => {}
+                            Ok(false) => {
+                                let err = interp.create_type_error("Cannot set property");
+                                return Completion::Throw(err);
+                            }
+                            Err(err) => return Completion::Throw(err),
+                        }
                     }
                     Completion::Normal(JsValue::Undefined)
                 },
@@ -814,11 +818,14 @@ impl Interpreter {
                         }
                     } else {
                         // Set(this, "constructor", v, true)
-                        if let Some(od) = interp.get_object_cell(this_id)
-                            && !od.borrow_mut().set_property_value("constructor", v)
-                        {
-                            let err = interp.create_type_error("Cannot set property constructor");
-                            return Completion::Throw(err);
+                        match interp.set_object_property(this_id, "constructor", v, this) {
+                            Ok(true) => {}
+                            Ok(false) => {
+                                let err =
+                                    interp.create_type_error("Cannot set property constructor");
+                                return Completion::Throw(err);
+                            }
+                            Err(err) => return Completion::Throw(err),
                         }
                     }
                     Completion::Normal(JsValue::Undefined)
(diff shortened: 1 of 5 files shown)


pmatos:
Validation environment note: the first unqualified full-suite run in the workspace timezone (`Europe/Berlin`) reported 22 baseline regressions, all in unrelated Intl DateTimeFormat/Temporal default-time-zone output. The branch started exactly at `origin/main`, and its diff contains no Intl/Temporal changes. Re-running the listed failures under `TZ=UTC` passed 20/20, and the complete UTC run passed 99,568/99,568 with zero regressions. I therefore used the UTC result as the baseline-comparable gate and left README.md unchanged at its existing 100.00% count.

PR #343, opened 2026-07-20
fix(node-shim): align TextDecoder encoding labels

### LINKED ISSUE #339: node-shim TextDecoder accepts non-standard unhyphenated encoding aliases (utf16le/ucs2/utf16be)
Deferred from #320 (zod harness): the shared `TextDecoder` shim accepts non-standard, unhyphenated encoding labels that real Node/Web `TextDecoder` rejects.

Original feedback by @chatgpt-codex-connector on PR #320 (https://github.com/pmatos/jsse/pull/320), posted after merge and never triaged:

> **Reject unsupported TextDecoder aliases**
>
> When bundled code uses Buffer-style labels such as `utf16le`, native Node's `TextDecoder` throws a `RangeError` (it accepts `utf-16le`/`ucs-2`, not the unhyphenated Buffer alias), but the jsse shim decodes successfully. That makes the same library bundle take different paths under the Node reference and jsse, so tests that verify Node/Web rejection of unsupported labels will fail in the host shim before exercising engine behavior; the other non-standard labels added alongside it have the same problem.

**Context:** `scripts/node-buffer-shim.js`, `normalizeTextDecoderEncoding` (~line 51). It maps both hyphenated and unhyphenated variants to the same encoding:

```js
case "ucs2":
case "ucs-2":
case "utf16le":
case "utf-16le":
  return "utf16le";
case "utf16be":
case "utf-16be":
  return "utf16be";
```

Per the WHATWG Encoding Standard's label table (which both Node's and browsers' `TextDecoder` implement), `utf16le`/`utf16be` (no hyphen) are not valid labels — only `utf-16le`/`utf-16` (and `ucs-2`/`csunicode`/`iso-10646-ucs-2`/`unicode`/`unicodefeff` for the LE case) are. Node throws `RangeError: The "encoding" argument must be of type string. Received an invalid encoding` (well, more precisely `RangeError [ERR_INVALID_ARG_VALUE]` / `RangeError: Unknown encoding: <label>` depending on version) for the unhyphenated forms. jsse's shim silently accepts them, so a bundle that probes `new TextDecoder("utf16le")` (or any test asserting the constructor throws for an unsupported label) takes a different path on jsse than on Node.

**Why deferred (scope-creep):** PR #320 wires up the zod normal+jitless corpus, which does not construct `TextDecoder` with any of the unhyphenated aliases, so the corpus's 2,184-case Node cross-check is unaffected either way. The finding is a real harness-fidelity gap but orthogonal to that PR's job.

**Suggested fix:** align `normalizeTextDecoderEncoding` with the WHATWG Encoding Standard label table — accept only the labels Node/browsers accept (`utf-8`, `utf-16le`, `utf-16`/`utf-16be` note: `utf-16` alone means LE, `unicode-1-1-utf-8`, etc. per the spec's table) and throw `RangeError` for Buffer-style unhyphenated aliases (`utf16le`, `ucs2`, `utf16be`) that are valid for `Buffer`/`normalizeEncoding` but not for `TextDecoder`.

_Filed following triage of review findings from the recent Node-compat library test-suite additions (js-md5 #322/#323, qs #322, uglify-js #321, zod #320); this was the one finding that slipped through untriaged because it was posted after PR #320 had already merged._

pmatos:
## Summary

- reject Buffer-only TextDecoder labels that Node and browsers reject
- accept the WHATWG aliases for the decoder families already implemented by the shim
- add a dedicated fixture cross-checked against native Node

## Testing

- `./scripts/run-shim-fixtures.sh` (266 Buffer assertions and 28 label assertions pass on both jsse and Node)
- `cargo fmt --check`
- `cargo clippy`
- `cargo build --release`
- `cargo test --release` (305 unit tests and smoke oracle pass)
- `uv run python scripts/run-test262.py` (completed 99,568 scenarios; 99,546 passed, with 22 unrelated pre-existing Intl/Temporal baseline failures and 323 new passes; engine and dependency inputs are identical to origin/main)

Closes #339

### DIFF
--- a/scripts/node-buffer-shim.js
+++ b/scripts/node-buffer-shim.js
@@ -45,27 +45,45 @@
     }
   }
 
-  // TextDecoder follows the Encoding Standard's label table rather than
-  // Buffer's encoding vocabulary. Keep this separate so accepting utf-16be
-  // here does not make Buffer.isEncoding("utf-16be") incorrectly return true.
+  // TextDecoder uses Encoding Standard labels for the decoder families below;
+  // Buffer has a different vocabulary. Keep normalization separate so
+  // accepting utf-16be does not make Buffer.isEncoding("utf-16be") return true.
   function normalizeTextDecoderEncoding(label) {
     switch (String(label).trim().toLowerCase()) {
-      case "utf8":
+      case "unicode-1-1-utf-8":
+      case "unicode11utf8":
+      case "unicode20utf8":
       case "utf-8":
+      case "utf8":
+      case "x-unicode20utf8":
         return "utf8";
+      case "ansi_x3.4-1968":
       case "ascii":
-      case "binary":
+      case "cp1252":
+      case "cp819":
+      case "csisolatin1":
+      case "ibm819":
       case "iso-8859-1":
+      case "iso-ir-100":
+      case "iso8859-1":
+      case "iso88591":
+      case "iso_8859-1":
+      case "iso_8859-1:1987":
+      case "l1":
       case "latin1":
       case "us-ascii":
       case "windows-1252":
+      case "x-cp1252":
         return "windows1252";
-      case "ucs2":
+      case "csunicode":
+      case "iso-10646-ucs-2":
       case "ucs-2":
-      case "utf16le":
+      case "unicode":
+      case "unicodefeff":
+      case "utf-16":
       case "utf-16le":
         return "utf16le";
-      case "utf16be":
+      case "unicodefffe":
       case "utf-16be":
         return "utf16be";
       default:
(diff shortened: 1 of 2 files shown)


pmatos:
Autonomous-run decision record:

- I rejected `binary` along with the three unhyphenated labels because Node rejects it for `TextDecoder` and it is the same Buffer-vocabulary leak. I also added the complete WHATWG label set only for the four decoder families the shim already implements. Removing only the three title labels would leave the same fidelity gap; implementing every WHATWG decoder would be out of scope.
- The mandatory full test262 run completed with 99,546/99,568 passing, 323 new passes, and 22 baseline regressions in unrelated Intl/Temporal formatting. `git diff --quiet origin/main -- src Cargo.toml Cargo.lock build.rs` confirms this branch has no engine or dependency-input changes, and test262 never loads the shim. I therefore treated those failures as baseline/environment drift and proceeded; the scoped jsse/Node fixture is green (266/266 Buffer assertions and 28/28 label assertions on each engine).

Issue #73, opened 2026-04-16
Optimize function call overhead: Environment pooling and fast argument binding


pmatos:
## Description

Every function call creates a new `Environment` (heap-allocated `Rc<RefCell<HashMap<String, Binding>>>>`), binds `this`, creates an arguments object, and binds each parameter. For mandreel's 849 emulated C functions (each taking a single `sp` argument), this per-call overhead is substantial.

## Proposed optimizations

1. **Environment pooling**: reuse `Environment` objects from a freelist instead of allocating new ones. Clear and repopulate bindings instead of allocating a new HashMap.

2. **Pre-sized Environment**: for functions with known parameter counts, pre-allocate the HashMap with the right capacity to avoid rehashing.

3. **Skip arguments object creation**: most functions don't use `arguments`. Lazily create the arguments object only when `arguments` is actually referenced.

4. **Fast parameter binding**: for simple parameter patterns (just identifiers, no defaults/destructuring), bind parameters with a single HashMap insert instead of the full `bind_pattern` machinery.

## Motivation

Mandreel (#54) calls thousands of small functions in tight loops. Each `function _something(sp) { ... }` call creates an Environment, binds `sp`, creates an unused arguments object, etc. Reducing this fixed per-call cost would significantly improve throughput.

## References

- #54 (mandreel benchmark timeout — parent issue)
- PR #65 (interpreter fast paths — partial mitigation)

pmatos:
I audited the merged #158 slice and the current call path before continuing this autonomous run. #158 cached IsSimpleParameterList, but direct parameter binding, environment pre-sizing, and pooling remain outstanding; sloppy ordinary calls also still eagerly create an arguments object for Function.prototype.arguments.

For this continuation I am implementing a bounded pool that recycles an EnvRef only when Rc::strong_count proves the activation did not escape. Unconditional reuse was rejected because returned or nested closures, mapped arguments objects, generators, and async work can retain the Function Environment Record. To make pooling effective for the motivating one-parameter sloppy calls, the Annex B getter will receive an inline deferred argument payload and materialize and cache the mapped object only if observed. Direct binding and pre-sizing will apply to all function kinds, while pooling stays limited to completed ordinary synchronous calls.

Alternatives considered were direct binding and pre-sizing alone (safe but leaves the main allocation cost) and pooling with eager sloppy arguments (safe but ineffective for mandreel-style calls). The detailed design is in docs/specs/2026-07-20-function-call-environment-pooling-design.md.

[...]

PR #369, opened 2026-07-23
fix(interpreter): guard boxing/error constructors against [[Call]] this-mutation

### LINKED ISSUE #355: Repeated calls to a class method with a default 2nd param + fn.call() break after a native-function first call
## Repro

```js
class C {
  m(fn, thisArg = this) {
    return fn.call(thisArg, 1);
  }
}
const c = new C();
console.log("1st (Boolean):", c.m(Boolean));
console.log("2nd (arrow):", c.m(() => "ok"));
```

- **jsse**: prints `1st (Boolean): true`, then throws `TypeError: undefined is not a function` on the second call.
- **Node**: prints both lines fine (`1st (Boolean): true` / `2nd (arrow): ok`).

## Isolating the trigger

All three of the following are required together — removing any one of them makes the bug disappear:

1. `m` is a method on a **class** (not a plain function).
2. `m` has a **default value for its second parameter** (`thisArg = this`).
3. The body calls **`fn.call(...)`** (not `fn(...)` directly).

The *first* call must pass a **native/built-in function** (e.g. `Boolean`) as `fn` — that call itself succeeds. Any *subsequent* call to the same method then throws, regardless of what's passed the second time (even `Boolean` again). If the first call instead passes a user-defined closure, all subsequent calls keep working fine.

More repros showing the isolation:

```js
// class method + default param + .call(), Boolean then Boolean: SECOND call also throws
class C {
  m(fn, thisArg = this) { return fn.call(thisArg, 1); }
}
const c = new C();
console.log(c.m(Boolean));  // true
console.log(c.m(Boolean));  // throws: TypeError: undefined is not a function

// any ONE of these changes makes it work fine:
// - drop the default param:      m(fn, thisArg) { ... }   (call with explicit thisArg)
// - drop .call():                m(fn, thisArg = this) { return fn(1); }
// - make it a plain function:    function m(fn, thisArg = this) { return fn.call(thisArg, 1); }
```

## Suspected cause

This looks like a regression from the recent function-call environment-pooling work (#73, `perf(interpreter): pool function call environments`, merged as part of PR #352 / commit `b7196d1`). The pattern suggests the pooled call environment for a method with a default parameter isn't reset/returned correctly when the bound parameter holds a native function, corrupting the environment slot for the next invocation of the same method.

## How this was found

Discovered while building the css-tree Node-compat library harness (#298): css-tree's own `List` class has `some(fn, thisArg = this)`/`filter(fn, thisArg = this)` methods matching this exact shape, and its test suite calls `list.some(Boolean)` before `list.some(() => {})` in the same test — hitting this bug. Left as 2 known/tracked failures (`List > #some() > basic`, `List > #filter() > basic`) out of 16,727 in that harness; the rest of the suite is cross-checked green against Node.

## Environment

- Built from this branch (`jsse/issue298`) off recent `main` (includes #73 pooled call environments, #66 arena allocator, #164 property-key intern cache).

pmatos:
## Summary

- `Boolean`, `Number`, `String`, `Error`, the `NativeError` subtypes (`TypeError`, `RangeError`, etc.), and `AggregateError` all mutated an arbitrary `this` object's internal slots — and, worse, its `[[Prototype]]` — whenever `this` happened to be an object during a plain `[[Call]]` (e.g. `Boolean.call(obj, 1)`), because none of them checked that `NewTarget` was actually defined before treating `this` as the pre-created object `OrdinaryCreateFromConstructor` sets up for `[[Construct]]`.
- **This turned out to have nothing to do with the pooled call-environment work (#73/PR #352) that the issue suspected** — it reproduces with a bare `Boolean.call({}, 1)`, no class, no default parameter, no pooling involved. The reported repro (`List#some(fn, thisArg = this)` calling `fn.call(thisArg, 1)` with a native function like `Boolean` as `fn`) just happened to be the first place a native function's `this`-clobbering behavior became externally visible: it silently rewrote the receiver's `[[Prototype]]` to `Boolean.prototype`, breaking every later call through that receiver.
- Fix: guard each constructor's this-mutation branch on `interp.new_target.is_some()`, matching the pattern already used correctly by `String`'s prototype-setting branch and `BigInt`.

Fixes #355

## Test plan

- [x] `cargo test --release` — 340 unit tests passing
- [x] `uv run python scripts/run-custom-tests.py` — 9/9 passing
- [x] `uv run python scripts/run-test262.py -j 32` — 99,568/99,568 (100.00%), 0 regressions, 323 new passes
- [x] New regression test `test262-extra/constructor-call-must-not-mutate-this.js` — covers `Boolean`/`Number`/`String`/`Error`+subtypes/`AggregateError` called via `.call()` with an arbitrary `this`, plus the exact issue repro; verified it fails on the pre-fix binary and passes after the fix
- [x] Manual repro confirms the bug is fixed

### DIFF
--- a/src/interpreter/builtins/mod.rs
+++ b/src/interpreter/builtins/mod.rs
@@ -566,7 +566,12 @@ impl Interpreter {
                         };
                     }
 
-                    if let JsValue::Object(o) = this {
+                    // §20.5.1.1 never reuses `this` — it always allocates via
+                    // OrdinaryCreateFromConstructor. A plain [[Call]] (e.g.
+                    // `Error.call(obj, msg)`) must not mutate an arbitrary `this`.
+                    if interp.new_target.is_some()
+                        && let JsValue::Object(o) = this
+                    {
                         if let Some(obj) = interp.get_object_cell(o.id) {
                             let mut o = obj.borrow_mut();
                             init_error!(o);
@@ -825,7 +830,12 @@ impl Interpreter {
                     1,
                     move |interp, this, args| {
                         let msg = args.first().cloned().unwrap_or(JsValue::Undefined);
-                        if let JsValue::Object(o) = this {
+                        // Only box into `this` when invoked via [[Construct]] — a plain
+                        // [[Call]] (e.g. `Test262Error.call(obj, msg)`) must not mutate
+                        // an arbitrary `this`.
+                        if interp.new_target.is_some()
+                            && let JsValue::Object(o) = this
+                        {
                             if let Some(obj) = interp.get_object(o.id) {
                                 let mut o = obj.borrow_mut();
                                 o.class_name = "Test262Error".to_string();
@@ -970,7 +980,12 @@ impl Interpreter {
                         };
                     }
 
-                    if let JsValue::Object(o) = this {
+                    // §20.5.6.1.1 never reuses `this` — it always allocates via
+                    // OrdinaryCreateFromConstructor. A plain [[Call]] (e.g.
+                    // `TypeError.call(obj, msg)`) must not mutate an arbitrary `this`.
+                    if interp.new_target.is_some()
+                        && let JsValue::Object(o) = this
+                    {
                         if let Some(obj) = interp.get_object(o.id) {
                             let mut o = obj.borrow_mut();
                             init_native_error!(o);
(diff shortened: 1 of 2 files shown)


pmatos:
Regarding the automated code-review comment above:

**Not an issue** — it's the `claude-review` check's own "No issues found" summary (that check run's conclusion is `success`). It raises no concern and requests no change, so there's nothing to fix, defer, or track.

_This assessment was made by two independent AI reviewers (Claude Sonnet 5 and Codex GPT-5.6-sol). If you disagree, please reply and we'll re-evaluate._

PR #396, opened 2026-07-25
perf(interpreter): collapse local binding writes to one lookup

### LINKED ISSUE #387: Perf: optimize global-scope statement execution (one-shot init call graphs)
## Description

Profiling #54 (mandreel benchmark) found that ~44.6% of its ~220s single-iteration runtime (98.3s) is spent inside a single top-level statement: `setupMandreel();`, called once at module-load/global-scope time before any benchmark iteration begins.

This is **not** slow global-variable-declaration machinery — that part is already fast. The same file also has 668 top-level statements of the form:
```js
var someFn__index__ = register_delegate(someFn);
```
executed one after another at global scope, and all 668 together take **9ms**. So writing to global bindings and calling a trivial function from global scope is cheap.

What's slow is that `setupMandreel()` fans out into a deep, broad, ONE-SHOT call graph — `startApp()` → `mandreelAppStart(...)` → ... — ultimately constructing a full Bullet-physics demo world. The mandreel file alone contains 788 distinct mangled C++ function symbols (`_Z...`), each a small function, mostly called once or a handful of times as part of this construction — not in a hot loop. Individual functions in this call graph are shaped like translated C (register-emulation locals `r0`/`r1`/.../`i7`, a `__label__` variable for control flow) rather than idiomatic JS — e.g. `__mandreel_internal_init` (also called from inside `runIteration`, see #54 for the ~77.6s figure there) is a single ~2,100-line function.

## Why this matters

Because each function in this graph runs only once (or a handful of times), none of the usual "hot function" optimizations apply — no bytecode-compilation payback (compile cost isn't amortized over repeat calls), no inline-cache warmup, no branch-prediction benefit from repetition. The cost is pure **baseline per-call and per-statement dispatch overhead**, multiplied by a very large number of *distinct* call sites/functions rather than a small number of call sites executed many times. This is a different performance problem than "make the hot loop faster" — it's "make every individual call/statement cheaper" — which benefits code that fans out broadly and shallowly (large one-shot init/construction code, as commonly produced by C++-to-JS compilers) as opposed to code that loops tightly.

## Suggested investigation

- Profile (e.g. via `perf` on the jsse binary, or targeted instrumentation) what actually dominates a single `call_function` invocation for these translated functions: environment/frame allocation, argument binding, local `var` slot creation, `Completion` enum construction/destruction, or something else.
- Check whether call-frame / environment allocation can be made cheaper for the common case (no closures captured, fixed arity, no `arguments` usage) — many of these register-emulated functions look like exactly this shape.
- Consider whether the existing prototype/property inline caches (#71) have any analogue applicable to *call* overhead itself (e.g. caching whether a callee needs a full environment record vs. a lightweight frame).
- This is explicitly **not** about the bytecode-VM's loop compilation (tracked separately in #67 and its follow-up, see the companion issue below) — it's about reducing the constant-factor cost of the interpreter's function-call and statement-execution paths in general, which would help one-shot init code AND make future bytecode-compiled hot loops cheaper to call into.

## Reproduction

Same recipe as #54: clone JetStream 3, run `Octane/mandreel.js`'s single `runIteration(0)`, with `Date.now()` markers around `setupMandreel()`. On the reference 61-core box (2026-07-24, jsse @ f0eb6a9): `setupMandreel()` alone took 98.27s wall while everything else in module load (parsing + 1,296 function/class declarations + 668 `register_delegate` calls) took 9ms.

## References
- #54 (mandreel benchmark timeout — parent context)
- #67 (bytecode compiler epic, closed — addressed loop compilation, not this)

pmatos:
## Summary

- update ordinary declarative local bindings through the same mutable map entry used for validation
- preserve the deferred global-object mirror path so proxy/re-entrant writes still happen without an active environment borrow
- add focused PutValue coverage and document the measured design tradeoffs

## Performance

- CPU-pinned adjacent A/B cold Mandreel runs against the exact `origin/main` binary used 14.7% to 20.9% less user CPU (15.4% median paired reduction across three pairs)
- a focused 20-million local-write check improved by 5.2% and 7.3% in reversed run order
- `perf` self-samples for `env_set` fell from 2.94% to 1.17%

## Validation

- `cargo fmt --check`
- `cargo clippy`
- `cargo build --release`
- `cargo test --release` (352 unit, 3 host-time-zone, smoke oracle)
- `./scripts/lint.sh`
- targeted assignment/function-code test262: 2,049/2,049
- full test262: 99,568/99,568, zero regressions
- focused `test262-extra` regression: 1/1
- custom tests: 11/11

Closes #387

### DIFF
--- a/docs/specs/2026-07-25-resolved-local-binding-write-design.md
+++ b/docs/specs/2026-07-25-resolved-local-binding-write-design.md
@@ -1,0 +1,76 @@
+# Resolved local binding write design
+
+## Goal
+
+Reduce repeated environment-map work when executing assignments in broad,
+one-shot call graphs without changing ECMAScript reference, binding, or global
+object semantics.
+
+The motivating Mandreel profile spends about 18% of cold execution in
+identifier resolution. For an assignment to an ordinary function-local
+binding, JSSE currently resolves the binding before evaluating the right-hand
+side and then rediscovers the same binding several times while checking
+mutability and writing the value.
+
+## Specification constraints
+
+`EvaluateCall`, `PrepareForOrdinaryCall`, and
+`FunctionDeclarationInstantiation` require a logically fresh activation and
+the prescribed parameter, `arguments`, var, lexical, and closure semantics.
+Assignment evaluation captures a Reference Record before evaluating the
+right-hand side; `PutValue` later writes through that captured record.
+
+The optimization may therefore retain and use the already-resolved
+Environment Record. It must not redirect the write by resolving the identifier
+again after the right-hand side runs. It must also preserve temporal dead zone,
+constant, immutable, function-name, global-object, import, `with`, and deleted
+binding behavior.
+
+## Design
+
+When `put_value_by_ref` receives `IdentifierRef::SpecificEnv`, first attempt a
+single mutable lookup in that exact Environment Record. If the record is not a
+global environment and the binding is still present:
+
+- initialized `var` and `let` bindings are written directly;
+- uninitialized lexical bindings report the existing TDZ error;
+- initialized `const` bindings report the existing assignment error;
+- function-name and immutable bindings retain their strict/sloppy behavior.
+
+If the record is global, the binding is indirect, or the binding disappeared
+while the right-hand side was evaluated, retain the existing general path.
+That path handles global object mirroring and proxy behavior, module imports,
+and unresolvable-reference semantics.
+
+This is a storage-level shortcut only. It does not omit an Environment Record,
+change reference capture timing, cache a mutable borrow across user code, or
+specialize based on function hotness.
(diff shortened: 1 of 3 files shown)


pmatos:
Autonomous design note: I first prototyped writing a resolved environment directly from the assignment dispatcher. That version regressed cold Mandreel CPU time from roughly 28–29 seconds to 34–36 seconds, so I discarded it. I chose the smaller `env_set` seam instead: one mutable binding-map probe for ordinary locals, with the global mirror path deliberately left deferred because it can invoke re-entrant object/proxy behavior. Slot-indexed frames could remove more hashing, but would require broader eval, closure, and declaration changes and is better handled as separate work.

PR #461, opened 2026-08-18
refactor(interpreter): extract pure eval hoisting analysis into a tested module

pmatos:
## Refactoring goal

Deepen the interpreter's most-churned file. An architecture audit (using the *improve-codebase-architecture* skill's process) flagged the `eval.rs` god file (17k lines, most-changed file in the last 200 commits) as the top hot spot. The most impactful, lowest-risk **deepening opportunity** there was a cluster of **pure AST static-semantics functions** that were buried inside `impl Interpreter` purely for namespacing:

- `collect_eval_var_names` / `collect_eval_var_names_from_stmt` / `collect_eval_function_decls` (in `eval.rs`)
- `unwrap_labeled_function` (in `exec.rs`)

They take no `self`, yet they encode a subtle, correctness-sensitive rule — descend into blocks / `if` / loops / `switch` / `try` / labels / `with`, but **never into nested function bodies** — and had **no test coverage**, because they could not be exercised without booting the whole interpreter. This is the classic *shallow-because-untestable* smell: the logic is pure, but its home hides it from tests (fails the deletion test only in the sense that its complexity was *hidden*, not *earned in place*).

### What changed
- New pure module `src/interpreter/hoisting.rs` with a small interface: `collect_var_names`, `collect_function_decls`, `unwrap_labeled_function`. No `Interpreter` state — so the traversal rules are unit-testable in isolation.
- `eval.rs` (−107 net) and `exec.rs` (−11 net) now call the module; ~120 lines of buried logic leave the god files.
- **No behavior change**: identical traversal, identical "keep the last function declaration of each name" de-duplication, identical consumption order for `EvalDeclarationInstantiation`.

### Before / after
```
Before: pure fns inside impl Interpreter (eval.rs 17k) + exec.rs  →  unreachable from cargo test
After:  interpreter::hoisting (pure, small interface)             →  7 unit tests, shared by eval + exec
```

## Tests that validate it (TDD)

Built test-first via the *tdd* skill in vertical slices (stub → red → implement → green), one seam at a time. Expected values are derived from the ECMAScript spec and cross-checked against `node` (authority order per CLAUDE.md: spec > test262 > node):

- **`collect_var_names`** — top-level and block-nested `var` hoist; control-flow / `try` traversal; **does NOT descend into nested function declarations** (`eval("var a; function g(){ var b; } a")` defines `a`, not `b`).
- **`unwrap_labeled_function`** — plain and (multiply-)labeled function declarations unwrap; non-functions and block-nested functions do not.
- **`collect_function_decls`** — de-duplication keeps the **last** declaration of each name (node: the second `f` wins), distinguished by arity; labeled top-level counts, block-nested does not.

### Validation results
- `cargo test --release`: **491 passed** (was 484), 0 failed — the +7 are the new module tests; the rest of the suite is unchanged.
- `./scripts/lint.sh`: rustfmt + `clippy -D warnings` clean.
- test262, 0 regressions: `language/eval-code` 454/454 · `language/statements/switch` 216/216 · `language/statements/function` 783/783 · `language/global-code` 75/75 · `annexB/language/function-code` 159/159.

## Notes for the reviewer
- This was an **autonomous** run. The skill's interactive "which candidate do you want?" step was resolved without an operator. The full candidate report (6 options, before/after diagrams, top recommendation) was generated to `/tmp/architecture-review-20260819-013053.html`.
- **Runner-up, deliberately deferred:** a confirmed latent bug — the typed-array element-coercion `to_number` at `src/interpreter/types.rs:3696` re-implements `helpers::to_number` and skips whitespace trimming and `0x`/`0o`/`0b` literals (reachable via `Object.defineProperty(typedarray, i, {value:"0x10"})`). It was **not** taken here because the fix is behavior-changing and MOP-invasive, which needs a full test262 sweep an operator-free run can't validate confidently. Filed as a follow-up issue.

🤖 Generated with Claude Code

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -13647,110 +13647,6 @@ impl Interpreter {
         self.dispose_resources(&lex_env, result.update_empty(JsValue::UNDEFINED))
     }
 
-    /// Collect top-level var-declared names from eval body (recursively into blocks, etc.)
-    fn collect_eval_var_names(stmts: &[Statement], names: &mut Vec<String>) {
-        for stmt in stmts {
-            Self::collect_eval_var_names_from_stmt(stmt, names);
-        }
-    }
-
-    fn collect_eval_var_names_from_stmt(stmt: &Statement, names: &mut Vec<String>) {
-        match stmt {
-            Statement::Variable(decl) if decl.kind == VarKind::Var => {
-                for d in &decl.declarations {
-                    d.pattern.bound_names(names);
-                }
-            }
-            Statement::Block(stmts) => {
-                for s in stmts {
-                    Self::collect_eval_var_names_from_stmt(s, names);
-                }
-            }
-            Statement::If(i) => {
-                Self::collect_eval_var_names_from_stmt(&i.consequent, names);
-                if let Some(alt) = &i.alternate {
-                    Self::collect_eval_var_names_from_stmt(alt, names);
-                }
-            }
-            Statement::While(w) => Self::collect_eval_var_names_from_stmt(&w.body, names),
-            Statement::DoWhile(d) => Self::collect_eval_var_names_from_stmt(&d.body, names),
-            Statement::For(f) => {
-                if let Some(ForInit::Variable(decl)) = &f.init
-                    && decl.kind == VarKind::Var
-                {
-                    for d in &decl.declarations {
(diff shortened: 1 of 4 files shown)


pmatos:
Follow-up for the deferred runner-up candidate (typed-array `to_number` divergence): filed as #462.

Issue #468, opened 2026-08-19
Bound the per-body IcStore: it pins every function body forever, keeping dynamic-function ASTs alive


pmatos:
`IcStore` (`src/interpreter/ic_store.rs`) keeps one `BodyIcStore` per function body, forever, and each one pins a clone of that body's `Rc<Vec<Statement>>` (`BodyIcStore._body`) so the `Rc::as_ptr` map key cannot ABA-collide. Nothing ever removes an entry: `IcStore` only does `index.get` / `index.insert` / `stores.push`.

This is the same retention class as #165 (the hoist-analysis cache), and the two are *co-dominant*: both caches independently pin the same body, so bounding either one alone does not let the AST be freed.

## Measured

`jsse` on a loop of N distinct `new Function` bodies (`var xI = I; { function gI(){} } return xI;`), each function object dropped right after its single call. Peak RSS, release build, temporary probe builds that skip one or both pins:

| N | baseline | hoist-cache pin skipped | IcStore pin skipped | both skipped |
|---|---|---|---|---|
| 5,000 | 50.5 MB | 48.4 MB | 50.3 MB | 36.4 MB |
| 20,000 | 111.2 MB | 104.3 MB | 111.6 MB | 38.8 MB |

Growth is ~3.7 KB per dynamic body while either pin remains, and flat when both are gone. With #165's bound landed, N=20,000 measures 107.0 MB — the ~4 MB the bounded hoist cache no longer retains — and still grows linearly, because `IcStore` holds every body.

## Why this is more than a capacity cap

`BodyStoreHandle` is a bare index into `IcStore::stores`, and live handles are held outside the store: in `Interpreter::current_ic_handle` and in the `prev` locals that `dispatch_body` / `exec_body` / `exec_eval_body` save and restore across recursion. Evicting a store therefore cannot just `remove` from the map and shrink the `Vec` — that would renumber or dangle handles belonging to bodies further up the call stack.

A workable shape (not the only one):

- Track in-flight use per store — increment where a handle is installed as `current_ic_handle`, decrement on restore — and make eviction skip any store with a non-zero count, so bodies on the current call stack are never evicted.
- Free evicted slots into a free list so `stores` does not grow monotonically, and tag handles with a generation so a stale handle is detected rather than silently aliasing a reused slot.
- Remove the map key and drop the pinned body in the same operation, keeping the ABA invariant that every surviving key names a live body.

An IC store is pure cache, so eviction only costs re-warming the slots.

## Context

[...]

Issue #72, opened 2026-04-16
Perf: cache hoisting analysis results on JsFunction


pmatos:
## Description

`collect_var_names_from_stmts` and `collect_annexb_function_names` walk the entire function body AST on every function invocation to collect var-hoisted names and Annex B block-scoped function names. For functions called thousands of times, this is redundant.

## Profiling data (from issue #55 callgrind)

| Function | % runtime | Instructions |
|----------|-----------|-------------|
| `hoist_vars_from_stmt` | 1.19% | 20.6M |
| `collect_annexb_function_names` | 0.51% | 8.7M |
| `hoist_pattern` | 0.19% | 3.3M |
| **Total** | **~2%** | **32.6M** |

## Proposed solution

Cache the results on `JsFunction::User`:
- Add `cached_var_names: Option<Rc<HashSet<String>>>`
- Add `cached_annexb: Option<Rc<(Vec<String>, HashSet<String>)>>`
- Compute on first invocation, reuse on subsequent calls

Since `body` is now `Rc<Vec<Statement>>` (from #55 PR), the cache key can use `Rc::as_ptr` for identity — if two JsFunctions share the same body Rc, they share the same cached analysis.

## Complexity note

The main challenge is borrow management: `JsFunction` is stored inside `Rc<RefCell<JsObjectData>>`, and mutating it to store the cache requires careful RefCell borrow handling during function call setup.

An alternative: store a global `HashMap<*const Vec<Statement>, Rc<HashSet<String>>>` on `Interpreter`, keyed by the body pointer.

PR #469, opened 2026-08-19
fix(interpreter): bound the per-body hoist-analysis cache

### LINKED ISSUE #165: Bound the per-body hoist_cache to prevent unbounded memory growth with many dynamic functions
Deferred from #157: the function-body hoisting-analysis cache (`hoist_cache`, added for #72) never evicts entries and pins each cached body via an owning `Rc`, so workloads that create many distinct dynamic function bodies grow memory without bound.

Original feedback by the Codex review bot on PR #157 (https://github.com/pmatos/jsse/pull/157#discussion_r3387476511):

> **Avoid pinning every dynamic function body forever**
>
> In programs that create and call many distinct dynamic functions, such as a loop around `new Function(src)()` or evaluated function expressions, this cached analysis stores a cloned `Rc` to every body in the interpreter-global `hoist_cache` and nothing ever removes entries (`rg hoist_cache` only finds initialization, lookup, and insertion). Those AST bodies could previously be freed once their function objects became unreachable, but now each first call retains the parsed body until the whole interpreter exits, so long-lived executions can grow memory without bound; consider a weak/evicting cache or tying cleanup to function/object lifetime.

**Context:** `src/interpreter/eval.rs` (lookup/insert, ~lines 1956–1973) and `src/interpreter/mod.rs` — `HoistAnalysis._body: Rc<Vec<Statement>>` pins the body so the `Rc::as_ptr` key can't ABA-collide; `hoist_cache: FxHashMap<*const Vec<Statement>, Rc<HoistAnalysis>>`. The cache only ever does `get` and `insert` — there is no `remove`/`clear`/`retain`/size cap anywhere.

**Why deferred:** diminishing-returns / unusual-workload — both AI reviewers confirmed the retention is real and a regression vs. pre-PR behavior, but it is not a correctness bug and the target workload (many distinct `new Function`/`eval` bodies in one long-lived interpreter) is uncommon for the project's test262-driven goal (each test runs in a fresh process). A correct bounded/evicting cache also interacts with the deliberate ABA-safety pinning invariant and warrants its own focused change plus a regression test, rather than expanding the scope of an otherwise self-contained perf PR.

**Suggested fix:** cap the cache size with simple eviction (LRU or capacity-based), keeping per-entry `_body` pinning so ABA safety still holds for entries that remain live; add a regression test that loops over many distinct `new Function` bodies and asserts the cache stays bounded.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 (DEFER) and Codex GPT-5.x (FIX) — evaluator disagreement defaults to DEFER._

pmatos:
## Summary

- The #72 hoisting-analysis memo (`Interpreter::hoist_cache`) was an unbounded `FxHashMap<*const Vec<Statement>, Rc<HoistAnalysis>>` whose entries each pin the analysed body's `Rc` (so the pointer key cannot ABA-collide). Nothing evicted, so every dynamic body a program ever ran — `new Function`, `eval` — was retained for the life of the interpreter.
- Replaced with a `HoistCache` in a new `src/interpreter/hoist_cache.rs`: capacity-bounded at 8,192 bodies, approximate-LRU, dropping the older half in one sweep when a new body would overflow. Entries are pure memoisation, so eviction only ever costs a re-walk of that body.
- `HoistAnalysis` moves into the same module and can only be constructed by the cache, which makes the ABA-safety invariant structural: a key exists only while the cache holds its body alive, and eviction removes the key and drops the pinned body in the same operation.
- Recency is tracked on lookup as well as insertion, so a hot body cannot be evicted ahead of a cold one — the `dispatch_body` fast path #72 added is preserved (`hoist_cache_still_reuses_one_analysis_across_repeated_calls` asserts the memo is still hit across repeated calls).

Fixes #165

## Scope: this bounds the cache; it does not on its own flatten RSS

`dispatch_body` populates two caches per body, and `IcStore` (`ic_store.rs`) pins the same `Rc<Vec<Statement>>` with no eviction of its own. The two pins are co-dominant: while either survives, the body's AST cannot be freed. Peak RSS on a loop of N distinct `new Function` bodies, each function object dropped right after its single call (release build; probe builds skip one or both pins):

| N | before | after (this PR) | hoist pin skipped | IcStore pin skipped | both skipped |
|---|---|---|---|---|---|
| 5,000 | 50.5 MB | 50.0 MB | 48.4 MB | 50.3 MB | 36.4 MB |
| 10,000 | 71.2 MB | 69.9 MB | — | — | — |
| 20,000 | 111.3 MB | 107.2 MB | 104.3 MB | 111.6 MB | 38.8 MB |
| 60,000 | — | 254.4 MB | — | — | — |

(The 60,000-body row is post-fix and well past the 8,192 cap: ~3.68 KB/body with no plateau, which is the `IcStore` pin, not this cache.) So: the cache itself is now bounded (and below the cap nothing changes by design — 5,000 bodies never evict), but total RSS still grows ~3.7 KB per dynamic body because `IcStore` holds every body. Bounding that needs handle-lifetime work — `BodyStoreHandle` is a bare index held in `current_ic_handle` and in `prev` locals across recursion — so it is filed separately as **#468** rather than widened into this PR. (The bytecode path in `dispatch_body` returns before the hoist-cache block, so this change is inert under `--bytecode` and needs no separate run there.) Wall-clock on the same workload is unchanged within noise on this (shared, loaded) machine; no throughput claim is made either way.

## Test plan

- [x] `cargo test --release` — 502 + 3 + 1 passing, 0 failed (9 new: 7 unit tests on the cache, 2 interpreter-level)
  - unit: memoisation by body identity, hit accounting, capacity bound over 1,000 churned bodies, eviction actually releases the pinned body (`Weak::upgrade` goes `None`), most-recently-used entry survives a sweep, re-insert of a cached body does not evict
  - interpreter-level: 10,192 distinct `new Function` bodies stay within `DEFAULT_CAPACITY` *and* fill it (guards the bound against passing vacuously), with the computed result asserted so eviction cannot silently change hoisting behaviour
- [x] `./scripts/lint.sh` — rustfmt + clippy `-D warnings` clean
- [x] test262 (`-j 32`): **99,568 / 99,568 (100.0%)**, 0 regressions
- [x] test262-extra: 144 / 144 (100.0%)
- [x] custom tests (`run-custom-tests.py`): 11 / 11
- [x] Manual repro (`new Function` churn at N = 5k/10k/20k) confirms the cache no longer grows past the cap; residual RSS growth attributed and filed as #468

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -2078,23 +2078,18 @@ impl Interpreter {
         // #72: cache the hoisting *name collection* for this function body,
         // keyed by the body's Rc pointer identity. The body Rc is stable across
         // function-object clones, so repeated calls reuse the analysis. ASTs are
-        // immutable post-parse, so the cache cannot go stale.
-        let key = Rc::as_ptr(&body.statements);
-        let analysis = match self.hoist_cache.get(&key) {
-            Some(a) => a.clone(),
+        // immutable post-parse, so the cache cannot go stale. The cache is
+        // capacity-bounded (#165), so a miss here can also be an eviction.
+        let analysis = match self.hoist_cache.get(&body.statements) {
+            Some(a) => a,
             None => {
                 let mut var_set = std::collections::HashSet::new();
                 Self::collect_var_names_from_stmts(body.as_slice(), &mut var_set);
                 let mut names = Vec::new();
                 let mut blocked = Vec::new();
                 Self::collect_annexb_function_names(body.as_slice(), &mut names, &mut blocked);
-                let a = Rc::new(HoistAnalysis {
-                    var_names: var_set.into_iter().collect(),
-                    annexb_names: names,
-                    _body: body.statements.clone(),
-                });
-                self.hoist_cache.insert(key, a.clone());
-                a
+                self.hoist_cache
+                    .insert(&body.statements, var_set.into_iter().collect(), names)
             }
         };
         let handle = self.ic_store.for_body(body);
--- a/src/interpreter/hoist_cache.rs
+++ b/src/interpreter/hoist_cache.rs
@@ -1,0 +1,260 @@
+//! Bounded memoisation of per-function-body hoisting analysis.
+//!
+//! `dispatch_body` runs the var/Annex-B *name collection* of
+//! `super::hoisting` once per function body and reuses it on later calls (#72),
+//! keyed by the identity of the body's statement `Rc`. Each entry pins that
+//! `Rc` so its address cannot be reused by a freed-then-reallocated body
+//! (ABA), which means an unbounded map would retain every body it has ever
+//! seen — including the short-lived bodies of `new Function` / `eval` (#165).
+//!
+//! (Bounding this cache is necessary but not sufficient to let such a body be
+//! freed: `super::ic_store` pins the same bodies and does not yet evict — #468.)
+//!
+//! The cache is therefore capacity-bounded with approximate-LRU eviction: when
+//! a new body would exceed `DEFAULT_CAPACITY`, the least-recently-used half of
+//! the entries is dropped in one sweep. Entries are pure memoisation, so eviction only ever
+//! costs a re-walk of the body. Eviction removes the map key and its pinned
+//! body together, so every surviving key still names a live body and the ABA
+//! invariant holds.
+
+use rustc_hash::FxHashMap;
+use std::rc::Rc;
+
+use crate::ast::Statement;
+
+/// Maximum number of memoised bodies. Large enough that a program with a fixed
+/// set of functions never evicts — the bound exists for body-churning workloads
+/// (`new Function` / `eval` in a loop). An entry costs a few hundred bytes plus
+/// the body it pins, so a full cache stays in the low megabytes for the small
+/// bodies such workloads produce.
+pub(crate) const DEFAULT_CAPACITY: usize = 8192;
+
+/// Cached output of the var/Annex-B hoisting *name collection* for a single
+/// function body (#72).
+///
+/// Only the raw name collection is cached. The Annex-B post-processing that
+/// inspects live env/parameter/lexical state still runs per call, and function
+/// declarations are never cached as values (fresh closures are built per call).
+/// The body `Rc` is pinned to prevent pointer (ABA) reuse from aliasing a
+/// freed body onto a fresh one with the same address.
+pub(crate) struct HoistAnalysis {
+    /// Deduped output of `collect_var_names_from_stmts`.
+    pub(crate) var_names: Vec<String>,
+    /// Raw `names` output of `collect_annexb_function_names`. (The companion
+    /// `blocked` accumulator is internal to the walk and discarded afterwards
+    /// by the original code, so it is intentionally not cached.)
+    pub(crate) annexb_names: Vec<String>,
+    /// Pins the body so its `Rc::as_ptr` key cannot be reused for another body.
+    _body: Rc<Vec<Statement>>,
+}
+
+struct Entry {
+    analysis: Rc<HoistAnalysis>,
+    /// Clock value of this entry's most recent lookup or insertion.
+    last_used: u64,
+}
+
+/// Capacity-bounded, approximate-LRU cache of `HoistAnalysis` keyed by body
+/// identity. The key `*const Vec<Statement>` is an identity only — never
+/// dereferenced. Entries cannot go stale: ASTs are immutable post-parse and
+/// each entry pins the body it is keyed by.
+pub(crate) struct HoistCache {
+    entries: FxHashMap<*const Vec<Statement>, Entry>,
+    capacity: usize,
+    clock: u64,
+    hits: u64,
+}
+
+impl HoistCache {
+    pub(crate) fn new() -> Self {
+        Self::with_capacity(DEFAULT_CAPACITY)
+    }
+
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            entries: FxHashMap::default(),
+            capacity: capacity.max(2),
+            clock: 0,
+            hits: 0,
+        }
+    }
+
+    /// Return the memoised analysis for `body`, marking it most-recently used.
+    pub(crate) fn get(&mut self, body: &Rc<Vec<Statement>>) -> Option<Rc<HoistAnalysis>> {
+        let key = Rc::as_ptr(body);
+        self.clock += 1;
+        let clock = self.clock;
+        let entry = self.entries.get_mut(&key)?;
+        entry.last_used = clock;
+        let analysis = entry.analysis.clone();
+        self.hits += 1;
+        Some(analysis)
+    }
+
+    /// Memoise the analysis of `body`, evicting the least-recently-used half of
+    /// the cache first if this entry would exceed the capacity.
+    pub(crate) fn insert(
+        &mut self,
+        body: &Rc<Vec<Statement>>,
+        var_names: Vec<String>,
+        annexb_names: Vec<String>,
+    ) -> Rc<HoistAnalysis> {
+        let key = Rc::as_ptr(body);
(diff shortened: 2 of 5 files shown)


pmatos:
Superseded by #472, which carries the same validated implementation on the issue branch required by the autonomous-run contract. No code-review findings were lost; #469 completed all checks successfully before replacement.

pmatos:
## Two-axis review of `0b81633` (base: `origin/main`)

Reviewed on two independent axes — **Standards** (does it conform to this repo's documented standards, plus a Fowler smell baseline) and **Spec** (does it faithfully implement #165) — by separate reviewers that did not see each other's findings. Line references below are against `0b81633`, before the follow-up commit.

### Standards

**(a) Documented-standard violations**

1. **`CONTEXT.md` domain language — `Body`: "_Avoid_: function body".** `hoist_cache.rs:1` ("per-function-body hoisting analysis") and `:4` ("once per function body") use the avoided term where the canonical noun is **Body**. Hard violation; small fix. The `HoistAnalysis` doc inherits the phrasing verbatim from `mod.rs`, so that part is pre-existing.
2. **"Use comments sparingly. Only comment complex code."** The ABA-pin invariant is stated five times (module doc twice, `_body`, the `HoistCache` doc, `evict_lru_half`). Keep it where the `_body` field needs the justification and cut the rest. `DEFAULT_CAPACITY`'s "stays in the low megabytes" is an unmeasured estimate that will rot — drop the sizing claim or measure it.

**(b) Smells (judgement calls)**

1. **Message Chain / abstraction mismatch with the analogue.** `get`/`insert` take `&Rc<Vec<Statement>>`, so the call site reaches inside `Body` two lines from the store that doesn't: `self.hoist_cache.get(&body.statements)` vs `self.ic_store.for_body(body)`. Take `&Body` in both and key internally, as `IcStore::for_body` does. Matters for #468, which will want parallel interfaces.
2. **Test-only instrumentation on a production path, diverging from in-repo precedent.** `#[allow(dead_code)]` on the whitebox accessors; the closest analogue, `gc::GcPacer`, gates its accessors (`alloc_count`, `bytes_since_gc`, `is_requested`, …) with `#[cfg(test)]`. Switch to `#[cfg(test)]`, including `hits`/`self.hits += 1`, which exist solely for one assertion.
3. **Miss path does three hash lookups** (`get` miss, `contains_key`, `insert`) and splits the analysis shape across call site and cache. Both collectors are static `Self::` fns, so `get_or_insert_with` is borrow-checker-feasible and collapses the `match`.
4. **Nits:** `capacity.max(2)` is undocumented (it guards `select_nth_unstable` against an empty `ticks`); one comment line is 97 cols where the file wraps at ~78 (rustfmt won't fix comments).

**Not findings:** the pointer-keyed-pin duplication with `ic_store` is explicitly deferred to #468 by the diff itself; `FxHashMap` vs `HashMap` and the `mod`/`use` interleaving are inherited/consistent. `AGENTS.md`/`CLAUDE.md` "Source Layout" is **stale**, not violated — it already omits `ic_store.rs`, `ic.rs`, `hoisting.rs`, `property_map.rs`, `scheduler.rs`, `bytecode/`.

### Spec

1. **Partial — the issue's headline goal is not met (disclosed).** Title: *"Bound the per-body hoist_cache to **prevent unbounded memory growth with many dynamic functions**"*. The **Suggested fix** as literally specified ("cap the cache size with simple eviction… keeping per-entry `_body` pinning… add a regression test") is fully implemented, but memory still grows linearly because `IcStore.index` (`ic_store.rs:31`) pins the same Bodies and is co-dominant. Extending the PR's own table past its largest N: 2,000 → 38.8 MB; 20,000 → 107.1 MB; 60,000 → 254.3 MB, i.e. ~3.68 KB/body with no plateau above the 8,192 cap. The PR states this and defers to #468 — accurate, but #165's stated purpose is not closed by this diff alone.
2. **Bound and ABA safety — no finding.** `len() >= capacity && !contains_key` keeps `len() <= capacity` on the real `dispatch_body` path. If an outer frame still holds an evicted entry's `Rc<HoistAnalysis>`, `_body` keeps that `Rc` alive, so the address cannot be reused — no ABA violation, only a duplicate re-analysis. `_body` is module-private, so "constructible only by the cache" holds.
3. **Recency/cutoff — one minor finding.** `get` does bump recency; `clock` increments at every assignment, so `last_used` values are unique and the cutoff cannot tie. But `select_nth_unstable(mid)` + `retain(> cutoff)` keeps `n - n/2 - 1`, and at the `capacity.max(2)` floor that keeps **zero** entries, evicting the MRU — contradicting the PR's *"a hot body cannot be evicted ahead of a cold one"*. Test-only at `DEFAULT_CAPACITY = 8192`.
4. **Regression test — exists, non-vacuous.** 10,192 Bodies would fail the `<= DEFAULT_CAPACITY` assert if eviction never fired, and the `> DEFAULT_CAPACITY / 2` guard blocks the converse.
5. **Miss vs hit observability — no finding.** A miss always recomputes *and* inserts, so `exec_statements_cached` always receives `Some(..)`; its `None` branch is unreachable from `dispatch_body`. `var_names` comes from a `HashSet` (nondeterministic order), but `exec_env` is always a function scope there, so `is_global` is false and `bindings` is an unordered `HashMap` — order unobservable. `annexb_names` is AST-ordered.
6. **Minor scope creep.** `hits()`, moving `HoistAnalysis` out of `mod.rs`, and the `ic_store.rs` comment weren't asked for. The `!contains_key` guard and `reinsert_of_a_cached_body_does_not_evict` cover a state `dispatch_body` cannot reach.

**Totals: Standards 6 findings (2 documented-standard, 4 judgement calls), worst = the `CONTEXT.md` `Body` term violation. Spec 3 findings (1 partial, 1 minor implementation, 1 minor scope creep), worst = the disclosed gap between #165's headline goal and what this diff can close alone.**

---

## Fixes applied (follow-up commit)

- **Spec 3 — MRU could be evicted at the capacity floor.** Cutoff is now `retain(last_used >= cutoff)`, keeping `ceil(len / 2)`, so the most-recently-used entry always survives and (given the floor of 2) a sweep still frees a slot. New test `a_sweep_at_the_capacity_floor_still_keeps_the_newest_entry` pins it — verified it **fails** with the old `>` cutoff.
- **Standards b1 + b3 — API mismatch and the three-lookup miss path.** The cache now exposes a single `get_or_insert_with(&Body, impl FnOnce(&[Statement]) -> (Vec<String>, Vec<String>))`, taking `&Body` like `IcStore::for_body` and keying internally. The call site no longer reaches into `body.statements`, the `contains_key` lookup is gone (a miss knows it missed), and with it the unreachable "insert a key that is already present" state and its test — which also closes half of Spec 6.
- **Standards b2 — test-only instrumentation.** `len()`, `hits()`, the `hits` field, and its increment are all `#[cfg(test)]` now, matching `gc::GcPacer`. No instrumentation in release builds.
- **Standards a1, a2, b4 — docs.** Reworded to **Body** per `CONTEXT.md`; the ABA-pin invariant is stated twice (the `_body` field, and the eviction sweep that must not break it) instead of five times; dropped the unmeasured "low megabytes" sizing claim; documented why the capacity floor is 2; rewrapped the over-long comment line.

**Deliberately not changed:** the `HoistAnalysis` move out of `mod.rs` (Spec 6) is the mechanism that makes "entries are built only by the cache" true, so it stays; the `ic_store.rs` cross-reference to #468 stays; Spec 1 needs #468, not a change here — the PR body carries the disclosure, now with the 60,000-body data point (254.4 MB, reproduced locally on the fixed build). `AGENTS.md`/`CLAUDE.md` "Source Layout" was left alone: it already omits five sibling modules, so listing only this one would be the odd entry.

**Re-validated after the fixes:** `cargo test --release` 501 + 3 + 1 green (8 cache tests + 2 interpreter-level), `./scripts/lint.sh` clean, test262 `**99,568 / 99,568 (100.0%)**, 0 regressions`, test262-extra `144 / 144`, custom tests `11 / 11`.

PR #472, opened 2026-08-19
fix(interpreter): bound the per-body hoist-analysis cache

### LINKED ISSUE #165: Bound the per-body hoist_cache to prevent unbounded memory growth with many dynamic functions

pmatos:
## Summary

- replace the unbounded per-Body hoisting-analysis map with an 8,192-entry approximate-LRU cache
- evict the colder half in one amortized sweep while preserving the per-entry Body pin that prevents pointer ABA collisions
- encapsulate cache construction so keys, recency metadata, and pinned analyses are updated together
- add focused eviction, pin-release, recency, and memoization tests plus dynamic Function integration coverage
- document the selected design and note that the separately tracked IcStore retention in #468 remains out of scope

## Validation

- cargo fmt --check
- cargo clippy
- ./scripts/lint.sh
- cargo build --release
- cargo test --release (502 passed, 1 ignored; integration tests 3 passed; smoke oracle 1 passed)
- Function test262 slice: 893 / 893
- function-statement test262 slice: 783 / 783
- test262-extra: 144 / 144
- custom tests: 11 / 11
- full test262: 99,568 / 99,568, zero regressions

Closes #165

### DIFF
--- a/docs/specs/2026-07-20-bounded-hoist-cache-design.md
+++ b/docs/specs/2026-07-20-bounded-hoist-cache-design.md
@@ -18,32 +18,39 @@ Body must transparently fall back to collecting the same names again.
 1. Store weak Body references and discard dead entries. This follows object
    lifetime closely, but it makes pointer reuse and entry validation more subtle
    and still permits unbounded growth while many distinct functions remain live.
-2. Clear the entire map when it reaches a fixed limit, matching the compiled
-   RegExp cache. This is very small, but a single miss discards every useful
-   analysis and can cause avoidable re-analysis cliffs.
-3. Keep a fixed-capacity map plus insertion-order queue and evict one oldest
-   entry on a full-cache miss. This is selected: lookup and insertion remain
-   O(1), eviction is deterministic, and no dependency or hit-path scan is added.
+2. Keep a fixed-capacity map plus insertion-order queue and evict one oldest
+   entry on a full-cache miss. This keeps lookup and insertion O(1), but a scan
+   across more Bodies than the capacity repeatedly evicts hot analyses along
+   with cold ones.
+3. Track a recency tick per entry and discard the least-recently-used half in
+   one sweep when the cache fills. This is selected: hits preserve hot analyses,
+   and bulk eviction amortizes the O(n) selection pass across many insertions.
 
 ## Design
 
 Introduce a small `HoistCache` wrapper containing the existing pointer-keyed map
-and a FIFO queue of its keys. Its capacity is 256 entries, consistent with the
-existing compiled RegExp cache bound.
+and a monotonic recency clock. Its capacity is 8,192 entries: large enough to
+avoid churn for ordinary programs with many static Bodies while still placing a
+fixed upper bound on body-churning `Function` and `eval` workloads.
 
-On a hit, the wrapper clones and returns the cached analysis without changing
-order. On a miss at capacity, it removes the oldest key from both collections
-before inserting the new analysis. Every retained entry continues to own its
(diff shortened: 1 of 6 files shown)


pmatos:
Post-merge review of this change, recorded here since it landed via this PR (my copy of the same first commit was on #469, which was closed unmerged — the review comment lives there and is easy to miss).

A two-axis review (Standards / Spec) plus a four-angle cleanup pass over the merged code found one real defect and several conformance items:

- **Defect:** `evict_lru_half`'s `retain(last_used > cutoff)` keeps `len - len/2 - 1` entries, which at the `capacity.max(2)` floor is zero — a sweep evicted the most-recently-used entry along with the oldest, contradicting the design note's "hits preserve hot analyses".
- The cache took its collection as a closure, so nothing bound a Body's key to the analysis stored under it; `Rc::as_ptr(&body.statements)` and its ABA contract were duplicated across both Body-keyed tables; whitebox accessors used `#[allow(dead_code)]` where the repo's precedent (`gc::GcPacer`) is `#[cfg(test)]`; and the module doc used "function body", which `CONTEXT.md` lists as an avoided term for **Body**.

Fixes are in #477. Reports: [two-axis review](https://github.com/pmatos/jsse/pull/469#issuecomment-5341895130); the cleanup pass's wider proposals (merging the two Body-keyed tables, or hanging per-Body data off the Body) are recorded on #468.

PR #485, opened 2026-08-24
fix(test262): collect extra module tests

### LINKED ISSUE #471: test262-extra/*.mjs files are collected by no test runner
## Problem

17 `.mjs` files in `test262-extra/` have never been executed by any runner, in CI or locally.

`scripts/run-test262.py`'s `find_tests()` collects `*.js` only:

```python
if path.is_file() and path.suffix == ".js":
    tests.append(path)
elif path.is_dir():
    tests.extend(f for f in sorted(path.rglob("*.js")) if not _is_fixture(f))
```

CI's only test262-extra steps are `run-test262.py … test262-extra/` (normal and `--bytecode`), so the `.mjs` files are silently skipped — an explicit `.mjs` path argument is dropped too, not just directory globs. `run-custom-tests.py` *does* collect `*.mjs`, but it is only ever pointed at `tests/`, and pointing it at `test262-extra/` would fail on the 76 `.js` files there because it prepends no harness.

Net effect: `test262-extra/` reports 145/145 while 17 files sit unexecuted. The affected set is the whole source-phase-imports corpus added in #220 plus a few others:

```
source-phase-import-eval-error.mjs        source-phase-reexport.mjs
source-phase-import-module-source.mjs     source-phase-shallow-load.mjs
source-phase-static-shallow.mjs           (+ 12 *-dep.mjs dependency modules)
```

## Why it stayed hidden

The count in CI goes up as `.js` extras are added, so a green `145/145` looks like full coverage. Nothing reports the skip.

## Impact

Real, but latent rather than active: I ran the 5 top-level `source-phase-*.mjs` files directly while working #222 and all 5 pass today. The risk is that they can regress silently — which nearly mattered for #222, since `source-phase-shallow-load.mjs` and `source-phase-static-shallow.mjs` guard the shallow-loading invariant that a host-resolver change can break.

## Options

1. **Rename the tests to `.js` with `flags: [module]`** — what #222's new test does. The runner already handles module mode (`compute_scenarios` → `[(file, "module")]`, `JsseAdapter.build_command` passes `--module`). Needs the 12 `-dep.mjs` files renamed to test262's `_FIXTURE` convention so `_is_fixture()` skips them, and their import specifiers updated.
2. **Teach `find_tests` to collect `.mjs`** — smaller diff, but `_is_fixture()` only recognises `_FIXTURE.{js,mjs}`, so the 12 `-dep.mjs` dependency modules would be collected as tests and fail. Would need `-dep.mjs` added to the fixture predicate, entrenching a jsse-local convention alongside test262's.
3. **Fail loudly on uncollected files** — have the runner error when a target directory contains `.mjs` files it will not run, so the gap cannot recur silently. Complements either of the above.

Option 1 plus 3 looks best: it converges `test262-extra/` on one collection rule and test262's own fixture naming, and makes a future omission noisy.

## Context

Found while investigating #222 (PR #470), which is why that PR's regression test is `.js` with `flags: [module]` rather than joining the `.mjs` group. Deliberately not fixed there — it is a runner change, orthogonal to the module-resolver fix.

pmatos:
## Summary

- convert seven executable `test262-extra` module tests to `.js` with `flags: [module]`
- rename eleven imported modules to the test262 `_FIXTURE.mjs` convention and update their specifiers
- reject non-fixture `.mjs` files that the test262 runner would otherwise omit silently
- cover explicit-file, directory, and fixture collection behavior in runner tests

## Validation

- `uv run python -m unittest scripts/test_run_test262.py`
- `cargo fmt --check`
- `cargo clippy`
- `cargo build --release`
- `cargo test --release`
- `./scripts/lint.sh`
- `uv run python scripts/run-test262.py --fail-on-failures test262-extra/` (172/172)
- `uv run python scripts/run-test262.py --bytecode --fail-on-failures test262-extra/` (172/172)
- `uv run python scripts/run-test262.py` (99,568/99,568; zero regressions)

Closes #471


pmatos:
## Code review pass (`/code-review --fix`, level: high)

Automated review of this PR found 7 findings; 6 applied as `569e767`, 1 reported only.

### Applied

1. **Guard was unscoped over the read-only `test262/` submodule** — `run-test262.py test262/tools/` aborted at collection with `non-fixture .mjs files would not be collected: test262/tools/regexp-generator/{header,index}.mjs`, and the remedy it printed (rename the files) is forbidden by CLAUDE.md rule 1. `_raise_for_uncollected_mjs` now takes `test262_dir` and exempts anything resolving inside the submodule.
2. **Dead `_raise_for_uncollected_mjs(default_dirs)` in the default branch** — the default corpus lives entirely inside the submodule, so once scoped the call can never fire, while costing a second full `rglob` over ~52k files on every default/`--sample` run. Removed.
3. **`run-custom-tests.py` fixture predicate went dead** — the `-dep.mjs` → `_FIXTURE.mjs` rename left `p.name.endswith("-dep.mjs")` matching nothing, so `run-custom-tests.py test262-extra` began collecting fixtures that throw by design (`source-phase-import-eval-error_FIXTURE.mjs`, `source-phase-shallow-target_FIXTURE.mjs`) as standalone failing tests. Replaced with a `_is_fixture()` matching run-test262.py's convention.
4. **No test pinned the guard *not* firing** — added `test_test262_submodule_mjs_does_not_block_collection`; finding 1 had shipped green through this file.
5. Redundant inner `sorted()` in `_uncollected_mjs` (caller re-sorts through a set) — dropped.
6. Trailing blank line in the design doc (`end-of-file-fixer` would rewrite it) — stripped; Design/Validation sections updated to record the submodule exemption and the run-custom-tests.py alignment.

### Reported, not fixed

7. Explicitly selecting a `_FIXTURE.mjs` path passes the guard but collects nothing, exiting with a bare `No tests found.` rather than explaining the fixture convention. **Deliberately skipped** — the fix is a UX decision on the existing `find_tests` contract, outside this diff's intent. Filing a follow-up is the reviewer's call.

### Verification after fixes

| Gate | Result |
|---|---|
| `python -m unittest discover -s scripts` | 15 tests OK |
| `run-test262.py test262-extra/ --fail-on-failures` | 95 files / 172 scenarios, 172/172 pass (also green under `--bytecode`) |
| `run-custom-tests.py` | 12/12 |
| `run-test262.py --sample 0.002` (exercises rewritten default branch) | 1,740 files, 3,352/3,352 pass |
| `./scripts/lint.sh` | rustfmt + clippy OK |
| `ruff format --check` | clean |

Discriminating checks, run independently of the review agent:

- `run-test262.py test262/tools/` now **collects and runs** (100 files) instead of aborting — exemption works.
- A stray `test262-extra/zz-stray-check.mjs` still exits with the guidance message — the guard is not merely disabled.

The 5 residual `ruff check` violations (S110, BLE001, PLW1510 ×2, PIE810) are all on lines this PR never touched; left alone.

pmatos:
### Addendum: verifying the `run-custom-tests.py` predicate dropped no tests

The predicate change is a *widening* (`-dep.mjs` → `_FIXTURE.js`/`_FIXTURE.mjs`), and a green `12/12` cannot distinguish "unchanged" from "a test was silently dropped" without a pre-change baseline. Settled by comparing both predicates over the current tree:

| root | files | collected (new) | collected (old) | newly skipped | no longer skipped |
|---|---|---|---|---|---|
| `tests/` (the default root) | 12 | 12 | 12 | — | — |
| `test262-extra/` | 106 | 95 | 106 | 11 | — |

- `tests/` is a **pure no-op** — no `*_FIXTURE.*` file exists there, so `12/12` was 12/12 before too. No test dropped.
- In `test262-extra/` the 11 newly-skipped files are exactly the intended `_FIXTURE.mjs` dependencies, and 95 matches `run-test262.py`'s collected count exactly — the two runners now agree.
- `no longer skipped` is empty in both, so nothing the old predicate excluded has started being collected.
- The broadened `_FIXTURE.js` arm is inert on the current tree (all 11 fixtures are `.mjs`); it exists so the two runners share one convention.

### Guard reachability

`.github/workflows/ci.yml:75,78` run `run-test262.py ... test262-extra/ --fail-on-failures` in both normal and bytecode modes, so the new guard is on the CI path — a future stray non-fixture `.mjs` in `test262-extra/` fails the build rather than being silently uncollected. That makes the #471 fix enforced, not just documented.

### Evidence caveat

Of the two `find_tests` branches, I re-ran the explicit-paths branch myself (`test262-extra/`, `test262/tools/`). The default no-paths branch — where the dead `_raise_for_uncollected_mjs(default_dirs)` call was deleted — is covered by the review agent's `--sample 0.002` run (1,740 files, 3,352/3,352), not by mine. The deletion removes a call whose only effect was raising, so a breakage there would be loud rather than silent; noting the asymmetry rather than papering over it.

pmatos:
## /simplify pass

Ran a four-angle simplification pass (reuse, simplification, efficiency, altitude) over this PR's diff. Applied in 9baaa0c.

### Applied

**Reuse — harness assertions.** The converted module tests gained `harness/assert.js` + `sta.js` preloads as a direct result of this PR (moving from bare `.mjs` under `run-custom-tests.py` to `.js` with `flags: [module]` under `run-test262.py`, which passes them via `--prelude` on the `is_module` path). Six files still hand-rolled what the harness now provides — five of them throwing bare `Error` rather than `Test262Error`. Swapped for `assert`, `assert.sameValue`, `assert.notSameValue`, `assert.throws`, per CLAUDE.md's "follow the exact same patterns of test262 tests". `SetterThatIgnoresPrototypeProperties-module-namespace-set.js`'s local `assertTypeError` helper is now just `assert.throws`. No `includes:` frontmatter needed — all four live in `assert.js`/`sta.js`.

**Simplification.** `_uncollected_mjs()` collapsed from four branches to two lines; the trailing `return []` was dead because `rglob` on a nonexistent path already yields `[]` (verified on this repo's CPython 3.13, not just in theory). Added `write_file()`/`frontmatter()` test helpers replacing repeated mkdir-parents + write_text plumbing across five call sites, and dropped the resulting dead `self.test262`.

**Guard message accuracy.** `_is_fixture` accepts `_FIXTURE.js` as well as `_FIXTURE.mjs`, but the error named only the latter — an author who already used the upstream `.js` spelling got no confirmation it was valid. Now `*_FIXTURE.js/.mjs`.

### Skipped (deliberate)

- **Renaming all `_FIXTURE.mjs` → `_FIXTURE.js`** to converge on upstream test262. A reviewer confirmed empirically that `.mjs` isn't load-bearing for fixtures, so this is a real option — but it reverses a decision the design doc records, and it changes which files the runner collects. That's a design call for this PR's author, not a simplification. Worth a follow-up issue if the convention split bothers us.
- **Dropping `rglob("*.mjs")` from `run-custom-tests.py`.** Would narrow what `tests/` collects — behavior change, and `tests/` has no `.mjs` today anyway.
- **Extracting a shared `_is_fixture`.** All three reviewers independently landed on keeping the duplication: there's no `scripts/` package, no cross-imports between any two scripts there, and both runners are hyphen-named (not importable without `importlib` gymnastics). A new module for a two-line predicate costs more than it saves.
- **Efficiency: nothing actionable.** Measured — the guard's extra walk is 0.4 ms on CI's actual input (`test262-extra/`), and `resolve()` fires zero times in practice since fixtures are filtered before it runs. The guard is also gated behind `if paths:`, so the full-corpus run and the mutation-testing oracle never invoke it.

### Verification

All baselines re-run after the edits and identical to before:

| check | result |
|---|---|
| test262-extra (normal) | 172/172 |
| test262-extra (bytecode) | 172/172 |
| `run-custom-tests.py` | 12/12 |
| `unittest discover -s scripts` | 15/15 |

Two extra checks, since a passing test isn't evidence a rewritten assertion still bites: mutating `assert.sameValue(ns.x, x, …)` → `42` and `assert.throws(TypeError, …)` → `RangeError` each turned their test red, so the new assertions are live rather than vacuous. Also confirmed a typo'd path still prints "No tests found." and exits 0 — the invariant the removed `return []` used to carry, which no test covers.

pmatos:
Two follow-ups to the pass above.

**Design doc synced** (080022d). My error-message change made the doc's *Error handling* section stale — it described the remedy as a `_FIXTURE.mjs` dependency only. Now names both spellings. The doc's `_FIXTURE.mjs` references elsewhere (lines 14, 19) describe the convention this PR deliberately adopted and are still accurate; left alone.

**Two notes for reviewers:**

*`assert.throws` is a slight semantic tightening.* The harness compares `thrown.constructor !== expectedErrorConstructor`, whereas the code it replaced used `caught instanceof TypeError` — so a TypeError *subclass* would have passed before and won't now. The tests pass, meaning the engine throws exact TypeErrors, and constructor-identity is the test262-conformant check, so this is a strictness improvement rather than a regression. Flagging it because the assertion changed shape, not just spelling. The async rejection checks correctly kept `instanceof`, since `assert.throws` can't wrap an `await`.

*Why no full test262 run.* CLAUDE.md asks for the full suite after implementation work. It isn't informative here: `_uncollected_mjs` (the only logic I changed in `run-test262.py`) is reachable only via `_raise_for_uncollected_mjs`, which is gated behind `if paths:` — the default full-corpus branch never calls it. The only other runner change is a message string. The 172-scenario `test262-extra` run in both engine modes exercises every path this commit actually touches.

PR #501, opened 2026-08-24
refactor(interpreter): narrow member LHS suspension channel

### LINKED ISSUE #466: Deepen: replace the eval_member_lhs_ref resume-value out-parameter with a narrow channel
## Deepening: replace the `eval_member_lhs_ref` resume-value out-parameter with a narrow channel

Split out of #456, whose "The real blocker" section flagged this. The
extraction in #456 landed without needing it, so this is a standalone
design cleanup rather than a blocker — but the wart is real and it will
matter for a later assignment/destructuring extraction.

### Corrected scope

#456 estimated "~33 sites in `eval.rs` plus more in `generator_transform.rs`".
That count conflates three unrelated things that share the `yield_val` name.
The actual out-parameter surface, verified against the tree at `ddb0166`:

- **1 signature** — `eval_member_lhs_ref(&mut self, .., yield_val: &mut Option<JsValue>)`
  (`src/interpreter/eval.rs:4705`). It is the only `&mut Option<JsValue>`
  parameter in the entire crate.
- **3 call sites** — `eval.rs:4813`, `4903`, `5145`, all in the
  assignment / destructuring-assignment code.

The remaining `yield_val` mentions are *not* this out-param:

- `Completion::Yield(yield_val)` pattern bindings and plain locals in
  `eval/generator_runtime.rs` (13) and `eval.rs:9178` — ordinary local
  names, no out-param involved.
- `ctx.new_temp_var("yield_val")` in `generator_transform.rs` (4) — a
  generated temp-variable *name string*.

### Why it is a wart

A yield can occur while evaluating the *reference* half of an assignment
target (`obj[yield x] = y`). Because the lref is pre-computed before the
RHS runs, the `Completion::Yield` cannot simply propagate from that
position, so the yielded value is smuggled back through the out-param and
re-raised by the caller. The result is that every caller of
`eval_member_lhs_ref` must remember to check `yield_val.is_some()` and
re-raise — a correctness obligation the type system does not enforce.

### Proposed change

Return the suspension in the value channel instead of an out-param, e.g.
have `eval_member_lhs_ref` return a two-variant enum
(`Ref(DestructLRef)` / `Suspended(JsValue)`), so the `?`/`match` on the
return value makes the yield path unmissable. Behavior-preserving.

### Verification

Full test262 run, per #456's precedent.

pmatos:
Summary:
- replace eval_member_lhs_ref's mutable yield out-parameter with an exhaustive Ref/Suspended return channel
- keep array iterator suspension bookkeeping local while making every helper caller handle suspension
- add focused coverage for base-expression and super computed-key suspension
- document the interface design and rejected broader channels

Validation:
- cargo fmt --check
- cargo clippy
- cargo build --release
- cargo test --release
- ./scripts/lint.sh
- uv run python scripts/run-custom-tests.py
- uv run python scripts/run-test262.py test262-extra/destructuring-assignment-member-target-suspension.js
- uv run python scripts/run-test262.py test262/test/language/expressions/assignment/dstr/ (640/640)
- uv run python scripts/run-test262.py (99,568/99,568; zero regressions)

Closes #466

### DIFF
--- a/docs/specs/2026-08-24-member-lhs-reference-suspension-channel-design.md
+++ b/docs/specs/2026-08-24-member-lhs-reference-suspension-channel-design.md
@@ -1,0 +1,107 @@
+# Member LHS reference suspension channel
+
+## Goal
+
+Remove the mutable resume-value out-parameter from
+`eval_member_lhs_ref` without changing assignment or destructuring behavior.
+The helper's interface must make a suspended member-reference evaluation
+impossible for any caller to overlook while remaining narrower than the
+interpreter's general `Completion` type.
+
+## Specification constraints
+
+ECMA-262 member-expression evaluation first evaluates the base and then, for a
+computed member, evaluates the key. Either evaluation can suspend at a
+`YieldExpression`. `EvaluatePropertyAccessWithExpressionKey` deliberately
+retains the raw key value: `ToPropertyKey` for `a[b] = c` is deferred until
+after the right-hand side has been evaluated.
+
+Destructuring assignment adds an ordering constraint. Iterator
+`AssignmentElement` and `AssignmentRestElement` evaluation computes a
+non-pattern target reference before stepping or draining the iterator.
+`KeyedDestructuringAssignmentEvaluation` likewise computes the target
+reference before reading the source property. A suspension at that point must
+therefore leave later iterator/property access and write-back unperformed.
+
+## Existing seam
+
+`eval_member_lhs_ref` pre-evaluates member, private, and `super` references for
+the assignment/destructuring module. It currently returns
+`Result<Option<DestructLRef>, JsValue>` while separately writing a yielded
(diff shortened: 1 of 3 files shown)


pmatos:
Autonomous design choice: the helper now returns a private two-variant MemberLhsRef channel (Ref(Option<DestructLRef>) / Suspended(JsValue)) inside the existing Result<_, JsValue> throw channel. This keeps non-member fallback behavior intact while making suspension exhaustive at every caller. I rejected Result<_, Completion> because it admits completion states the helper cannot produce, and a separate NotMember/Ref/Suspended enum because it expands the interface only to replace the Option callers already consume. Array destructuring retains its local yield state solely for shared iterator-close/replay bookkeeping. The full rationale is recorded in docs/specs/2026-08-24-member-lhs-reference-suspension-channel-design.md.

PR #510, opened 2026-08-24
fix(interpreter): resolve inline labelled continue targets

### LINKED ISSUE #496: Labelled `continue` is ignored in async for-of loops
## Summary

`continue <label>` targeting an outer `for-of` from a nested `for-of` inside an async function is treated as an unlabelled `continue` of the innermost loop, so the outer loop body runs to completion instead of advancing the outer iteration.

## Reproducer

```js
async function f(){
  outer:
  for (const x of [1,2]) {
    for (const y of [3,4]) {
      await null;
      continue outer;
    }
    console.log("not reached " + x);
  }
  console.log("done");
}
f();
```

| Engine | Output |
|---|---|
| jsse | `not reached 1` / `not reached 2` / `done` |
| node | `done` |

## Cause

Two parts:

1. `src/interpreter/eval.rs` (~8688, ~8697) selects the target loop with `for_of_stack.iter().rposition(|_| label.is_none())` — the predicate ignores the element entirely, so it is label-blind and always picks the innermost loop.
2. `transform_labeled_statement` in `src/interpreter/generator_transform.rs` (~2074-2097) registers only `break_targets`, so there is no continue target to resolve against.

`AsyncForOfLoopState` (added in PR #486) is the natural home for a `label` field to make the driver side resolvable.

## Spec

§14.7.5.7 / §13.x: a continue completion with a label propagates outward until it reaches the labelled iteration statement.

Found during a simplification review of PR #486.

pmatos:
## Summary

- carry transformed for-of label sets into active async loop state
- resolve inline labelled continue completions against the correct outer loop
- cover iterator unwinding when a non-suspending try surfaces continue after await

## Validation

- cargo fmt --check
- cargo clippy
- ./scripts/lint.sh
- cargo build --release
- cargo test --release
- uv run python scripts/run-custom-tests.py
- focused test262-extra regression: 2/2
- test262 for-of: 1,442/1,442
- full default test262: 99,568/99,568 (100.00%)

Closes #496

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -8824,7 +8824,9 @@ impl Interpreter {
                     // An inline statement can surface continue directly rather
                     // than through a LoopControl terminator. Route it through
                     // intervening finalizers before returning to the loop head.
-                    if let Some(pos) = for_of_stack.iter().rposition(|_| label.is_none()) {
+                    if let Some(pos) = for_of_stack.iter().rposition(|loop_state| {
+                        loop_state.matches_continue_target(label.as_deref())
+                    }) {
                         let loop_state = &for_of_stack[pos];
                         let target = LoopControlTarget {
                             target_state: loop_state.head_state,
@@ -9093,6 +9095,7 @@ impl Interpreter {
                 StateTerminator::ForOfInit {
                     ref iterable,
                     ref iter_var,
+                    ref label_set,
                     ref left,
                     head_state,
                     after_state: forinit_after,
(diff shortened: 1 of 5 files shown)


pmatos:
Investigation note: the issue’s bare `continue outer` reproducer was already green on current `main` because #486 added transformed labelled-continue targets and regression coverage. The driver-side label-blind path remained reachable when a non-suspending inline `try` surfaced `Completion::Continue(Some("outer"))` after an `await`; that deterministic variant still produced the reported unreachable output. This PR fixes and covers that remaining seam rather than duplicating the existing bare-case test.

Issue #502, opened 2026-08-24
Object destructuring source is not GC-rooted across computed-key user code


pmatos:
Found while reviewing #498 (issue #464).

## Summary

`destructure_object_assignment` holds the `ToObject` result in an untraced Rust
local (`src/interpreter/eval.rs:4644`). Every computed property key and every
source getter in the pattern runs arbitrary user code *after* that local is
taken, so a collection during that code frees the source object and the arena
slot is reused. Subsequent property reads from the pattern then observe
`undefined`.

## Repro

```js
function f() { $262.gc(); return "y"; }
var a, b;
({ x: a, [f()]: b } = { x: 1, y: 2 });
print("a=" + a + " b=" + b);
```

| engine | result |
| --- | --- |
| jsse | `a=1 b=undefined` |
| node (`f` without gc) | `a=1 b=2` |

Without the `$262.gc()` call jsse also prints `a=1 b=2`, so this is specifically
a rooting failure rather than a key-evaluation-order bug. A primitive source
(where `ToObject` allocates the wrapper) fails the same way.

## Why it was deferred out of #498

PR #498 roots the destructuring *target* side: the lRef base, its un-coerced
key, the super receiver, each rest value as collected, and the value awaiting
the write. That fix lands as a `gc_root_frame()` inside the per-element loop.

The source-side `obj_val` cannot use the same frame. It is taken before the
loop, and `destructure_object_assignment` has roughly six early returns between
the `to_object` call and the loop body. Rooting it at the point it is created
without restructuring the function to a single exit would leak a root on those
early-return paths and trip the leak assertions the PR added. The repair is a
single-exit restructure of that function, which is a larger change than #464
scopes.

## Suggested fix

Restructure `destructure_object_assignment` so the body after `to_object` runs
under one `gc_root_frame()` / `gc_unroot_frame()` pair with a single exit, then
`gc_root_value(&obj_val)` immediately after `to_object` succeeds.

## Test

Add a `test262-extra` case covering a computed key whose `toString` collects and
a getter that collects, for both an object source and a primitive source. The
existing `test262-extra/destructuring-target-value-gc-rooting.js` covers the
target side only.

PR #514, opened 2026-08-24
fix(interpreter): root computed member target base

### LINKED ISSUE #505: Member target base is not GC-rooted across computed-key user code
## Summary

`eval_member_lhs_ref` evaluates a member target's base into a Rust local and then
evaluates the computed key expression while that local is untraced. If the key
expression runs user code that forces a collection, the base is swept before the
`DestructLRef` that will hold it is ever constructed — so the eventual `PutValue`
writes to a dead or reused arena slot and the setter never runs.

This is the same failure class as #464, but it sits in the *producer* of the
reference rather than in its consumers. #498 roots everything a `DestructLRef`
holds immediately after `eval_member_lhs_ref` returns, which by construction
cannot cover the window *inside* that function.

## Reproduction

```js
var w = [];
var proto = { set y(v) { w.push(v); } };
var iter = {
  [Symbol.iterator]: function () {
    var taken = false;
    return { next: function () {
      if (taken) return { done: true };
      taken = true;
      return { value: "V", done: false };
    } };
  }
};
[ Object.create(proto)[ (function(){ $262.gc(); for (var i=0;i<200;i++){ ({a:i}); } return "y"; })() ] ] = iter;
print("writes=" + w.length + " value=" + w[0]);
```

| engine | result |
| --- | --- |
| jsse (this repro) | `writes=0 value=undefined` |
| jsse, with `$262.gc()` removed | `writes=1 value=V` |
| node | `writes=1 value=V` |

The object-destructuring form `({ p: Object.create(proto)[keyExpr()] } = src)` fails
identically. Removing the `$262.gc()` makes both pass, which isolates the cause to
collection rather than to evaluation order.

## Cause

`src/interpreter/eval.rs`, in `eval_member_lhs_ref`:

```rust
let base = match self.eval_expr(obj_expr, env) { ... };   // untraced Rust local

match prop {
    MemberProperty::Computed(key_expr) => {
        let raw_key = match self.eval_expr(key_expr, env) { ... };  // can collect `base`
        Ok(Some(DestructLRef::Member(base, raw_key)))
    }
    ...
}
```

`base` is live across `eval_expr(key_expr, env)` but is not in `gc_temp_roots`.

## Suggested fix

Self-contained inside `eval_member_lhs_ref`, no caller change: take a
`gc_root_frame()` after `base` is evaluated, `gc_root_value(&base)`, evaluate the
key, and `gc_unroot_frame` before returning. Because it is balanced within the
function, the callers' own frames and their pre-frame early-`break` paths are
unaffected.

Note that hoisting the *callers'* frames above the `eval_member_lhs_ref` call is
**not** an equivalent fix — the `Err(e) => break` and `yield_val.is_some() => break`
paths currently run before those frames open, and would then leak roots.

## Scope note

Found during the `/simplify` pass on #498 (altitude review). Deliberately not fixed
there: it is a behavior change, it is pre-existing rather than introduced by that
PR, and folding it into a simplification commit would ship it unreviewed.

Related: #464, #498, #502.

pmatos:
## Summary

- keep a computed member target's evaluated base GC-rooted while its key expression runs user code
- balance the local root frame before propagating normal, throw, or suspension results
- cover array/object destructuring under forced GC and verify abrupt key evaluation does not leak roots

## Validation

- `cargo fmt --check`
- `cargo clippy`
- `./scripts/lint.sh`
- `cargo build --release`
- `cargo test --release`
- `uv run python scripts/run-custom-tests.py`
- targeted test262: 17/17
- full test262: 99,568/99,568 (100.00%, zero regressions)

Closes #505

### DIFF
--- a/src/interpreter/eval.rs
+++ b/src/interpreter/eval.rs
@@ -4293,7 +4293,16 @@ impl Interpreter {
                 JsValue::string(JsString::from_str(name)),
             )))),
             MemberProperty::Computed(key_expr) => {
-                let raw_key = match self.eval_expr(key_expr, env) {
+                // The base becomes the Reference Record's [[Base]], so it must
+                // survive arbitrary user code while the computed key is
+                // evaluated. Callers cannot root it until this function has
+                // finished constructing the reference.
+                let gc_frame = self.gc_root_frame();
+                self.gc_root_value(&base);
+                let key_result = self.eval_expr(key_expr, env);
+                self.gc_unroot_frame(gc_frame);
+
+                let raw_key = match key_result {
                     Completion::Normal(v) => v,
                     Completion::Throw(e) => return Err(e),
                     Completion::Yield(v) => return Ok(MemberLhsRef::Suspended(v)),
--- a/src/interpreter/tests.rs
+++ b/src/interpreter/tests.rs
@@ -1518,6 +1518,24 @@ fn private_destructuring_releases_temp_roots_after_abrupt_completion() {
 }
 
 #[test]
+fn computed_member_destructuring_releases_temp_roots_after_abrupt_key_evaluation() {
+    // The member base is rooted while its computed key is evaluated. An abrupt
+    // key completion must release that scoped root before propagating the throw.
+    let interp = run_script(
+        r#"
+        function fail() { throw new Error("computed key"); }
+        try { [Object.create({})[fail()]] = [1]; } catch (e) {}
+        try { ({ item: Object.create({})[fail()] } = { item: 1 }); } catch (e) {}
+        "#,
+    );
+
+    assert!(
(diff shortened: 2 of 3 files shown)


pmatos:
Scope note from implementation: an inline, otherwise-unrooted RHS object in the object-destructuring variant can also be collected while the computed target key runs, which is a distinct source-object lifetime edge. Issue #505 and its reproduction isolate the target base with an independently rooted `src`, so this PR keeps that same isolation via `computedPropertySource`. I considered also rooting the object-destructuring source here, but that would broaden this focused producer fix and deserves separate review.

pmatos:
Follow-up to the scope note: the distinct object-destructuring source lifetime edge is already tracked as #502, so keeping this PR focused on the member target base preserves the intended split.

PR #547, opened 2026-09-01
chore(test262): update submodule to 7710052

pmatos:
## Summary

Moves the test262 pin from `f2d1435` (2026-07-13) to `7710052` (2026-08-30): 41 commits, 227 test files changed, +5,932 lines. Scenarios rise 99,568 → 99,907.

**Result: 99,620 / 99,907 (99.71%), 287 failures, 0 regressions.** Three consecutive full runs produced identical numbers.

## No regressions

Every one of the 287 failures is a test *this bump adds*. Verified per file rather than trusting the runner's regression count alone — no previously passing test fails.

The baseline grows by 373 rather than 381 new passes. The 8-entry difference is fully accounted for: three `array-rest-elision-invalid` dstr files were deleted upstream (6 scenarios), and one `Atomics/pause` entry was already stale on `main` — its test was deleted upstream *before* main's own pin, so `test262-pass.txt` on main lists a file that does not exist.

## Where the 287 failures are

| Area | Scenarios | Cause |
|---|---:|---|
| `Iterator.prototype.includes` | 80 | proposal not implemented |
| `Iterator.prototype.windows` | 70 | `iterator-chunking` not implemented |
| `Iterator.prototype.chunks` | 68 | `iterator-chunking` not implemented |
| `Iterator.prototype.join` | 32 | proposal not implemented |
| `Intl.Locale` info getters | 24 | `-u-rg-` region override ignored |
| dstr `obj-rest-before-comma` | 6 | trailing comma after rest not a SyntaxError |
| `Promise.try` avoids-wrap | 4 | new spec, returned promise must not be wrapped |
| module-level `using` | 2 | disposal at end of module evaluation missing |
| `import defer` + TLA cycle | 1 | cannot evaluate deferred module with async cycle dep |

**250 of 287** are the four unimplemented `Iterator.prototype` proposals — all four are `undefined` on the prototype. Not engine defects, just unbuilt features.

The remaining **37 are genuine gaps**, each confirmed by hand against Node:

- **`-u-rg-` region override ignored.** `new Intl.Locale("en-US-u-rg-gbzzzz").getHourCycles()` → `["h12"]`; Node → `["h23"]`. jsse's Intl data *is* region-sensitive (`en-US` vs `en-GB` match Node exactly), it just never uses the `rg` extension as the lookup region. Accounts for all 24.
- **Trailing comma after object rest accepted in assignment patterns.** `0, {...rest,} = {}` parses; Node throws. jsse already rejects it in *binding* patterns (`var {...rest,} = {}`), so the two paths disagree.
- **Module-level `using` never disposes** — module body completes with `disposed === false`.
- **`import defer` with an async cycle dependency** throws `TypeError: Cannot synchronously evaluate a module with top-level await`.
- **`Promise.try` wrapping** — note **Node fails this one too**; it is a very recent spec change neither engine implements yet.

## Baseline and README

`test262-pass.txt` is rolled forward so the 381 newly passing tests are protected against future regressions.

The README headline goes from `100.00%` to `99.71%`, stated plainly with the breakdown rather than scoped to hide the unimplemented proposals.

## Note

The tightened upstream `Iterator/prototype/{take,drop}/limit-rangeerror.js` caught a real engine bug while preparing this; that fix landed separately in #545 and is already on `main`.

https://claude.ai/code/session_01SWtdLjDrAqjE6YCDYaPvkH


pmatos:
Filed one issue per root cause; together they account for all 287 failures:

| Issue | Scenarios | Cause |
|---|---:|---|
| #548 | 138 | `Iterator.prototype.chunks` / `.windows` not implemented (`iterator-chunking`) |
| #549 | 80 | `Iterator.prototype.includes` not implemented |
| #550 | 32 | `Iterator.prototype.join` not implemented |
| #551 | 24 | `Intl.Locale` info getters ignore the `-u-rg-` region override |
| #552 | 6 | trailing comma after object rest accepted in assignment patterns |
| #553 | 4 | `Promise.try` wraps a returned promise (Node fails these too) |
| #554 | 2 | module top-level `using` never disposed |
| #555 | 1 | `import defer` with an async cycle dependency throws |

138 + 80 + 32 + 24 + 6 + 4 + 2 + 1 = **287**.

#548, #549 and #550 are unimplemented proposals — 250 of the 287, and the bulk of what stands between this pin and a 100% pass rate again. The other five are engine defects, each verified by hand against Node.

Issue #535, opened 2026-08-25
ci: claude-review fails on every PR — CLAUDE_CODE_OAUTH_TOKEN is not usable


pmatos:
The `Claude Code Review` workflow (`.github/workflows/claude-code-review.yml`) has failed
on **every** run since at least 2026-08-25, on unrelated branches, with an identical
authentication-shaped failure. It is not caused by any PR's contents.

## Evidence

Every recent run of the workflow, across two independent branches:

| created | conclusion | branch |
|---|---|---|
| 2026-08-25T22:20Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T22:03Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T21:32Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T20:59Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T20:45Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T20:21Z | failure | `sym/jsse/522-…` (PR #532) |
| 2026-08-25T19:55Z | failure | `sym/jsse/525-…` |

## Failure signature

Identical in every run. The SDK starts, initialises, and returns an error result on the
very first turn without doing any work:

```
{ "type": "system",  "subtype": "init", "message": "Claude Code initialized", "model": "claude-sonnet-5" }
{ "type": "result",  "subtype": "success", "is_error": true,
  "duration_ms": 593, "num_turns": 1, "total_cost_usd": 0, "modelUsage": {} }
##[error]Claude result reported subtype success with is_error:true (run did not complete successfully)
##[error]Action failed with error: Claude execution failed: result is_error:true
##[error]Process completed with exit code 1.
```

`total_cost_usd: 0`, `modelUsage: {}` and `num_turns: 1` after ~0.6 s mean no model call
was ever billed — the run is rejected before any inference. That is the shape of a
credential problem, not a prompt or repository problem. Marketplace install and plugin
install both succeed beforehand, so it is not a plugin-resolution failure.

## Likely cause

`.github/workflows/claude-code-review.yml:54` authenticates with:

```yaml
claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
```

The most probable causes, in order: the secret has **expired** (OAuth tokens are not
indefinite), it was rotated without the repo secret being updated, or the account behind
[...]

PR #565, opened 2026-09-01
fix(parser): reject trailing commas after object rest patterns

### LINKED ISSUE #552: Trailing comma after object rest accepted in assignment patterns
## Summary

A trailing comma after a rest element in an object **assignment** pattern must be a SyntaxError. jsse accepts it.

## Reproduce

```js
var rest;
0, {...rest,} = {}     // jsse: parses fine    node: SyntaxError: Rest element must be last element
```

The two parser paths disagree with each other — jsse already rejects the same construct in a **binding** pattern:

```js
var {...rest,} = {}    // jsse: SyntaxError (correct)
```

So the object *binding* pattern path enforces the rule and the object *assignment* pattern path does not.

## Impact

**6 failing scenarios across 3 files**, surfaced by the test262 bump to `7710052` (#547):

- `language/expressions/assignment/dstr/obj-rest-before-comma-invalid.js`
- `language/statements/for-in/dstr/obj-rest-before-comma-invalid.js`
- `language/statements/for-of/dstr/obj-rest-before-comma-invalid.js`

All three are `negative: { phase: parse, type: SyntaxError }` tests.

## Reproduce (suite)

```
uv run python scripts/run-test262.py test262/test/language/expressions/assignment/dstr/obj-rest-before-comma-invalid.js
```

## Notes

Upstream also deleted three `array-rest-elision-invalid` dstr files in the same bump, replacing them with this stricter `obj-rest-before-comma` coverage — so this area of the grammar was deliberately retested upstream.

pmatos:
Reject object assignment patterns whose final rest property is followed by a trailing comma, while preserving the same syntax in ordinary object literals.

The parser now carries the source-level comma shape on object expressions and checks it at both cover-grammar reinterpretation paths: assignment expressions and `for-in`/`for-of` or arrow parameter patterns. Parser regressions cover top-level, nested, loop, arrow, and nearby valid forms.

Validation:
- `./scripts/lint.sh`
- `cargo build --release`
- `cargo test --release`
- targeted test262: 4,771/4,771 passing, including all 6 issue scenarios
- full test262: 99,764/99,907 passing (99.86%), 0 regressions

Closes #552


pmatos:
This workspace was reused from a prior automated run, which left an uncommitted change to `src/interpreter/generator_transform.rs` (threading the object `trailing`-comma flag through instead of hardcoding it to `false` in two rewrite arms).

Judgment call: committed it (`6702332`) rather than discarding it, since it was already correctly implemented and validated.

Verified behavior: I compared jsse against Node for `{ ...rest, }` patterns (including forced through the generator-transform path via a `yield` in the pattern's value) and for object literals/patterns with trailing commas in generator bodies. Behavior is identical with and without this change — the trailing-comma-after-rest early error fires at parse time, before the generator transform ever runs, so this field isn't consulted for that diagnostic. This is a **consistency-only fix** (the `Expression::Array` arm in the same `rewrite_expr` already threads `*trailing`; `Expression::Object` now matches it), not a behavioral bug fix — the commit message overstates it as fixing a live defect, which I want to correct here for the record.

Validation: `cargo build --release`, `cargo test --release --bin jsse` (602 passed), `./scripts/lint.sh` (clean), and a targeted test262 slice (`language/expressions/object`, `language/expressions/assignment`, `language/expressions/generators`, `language/statements/generator` — 3648/3648 passing, 0 regressions vs. baseline).

Alternative considered: revert the uncommitted change and leave the inconsistency in place. Rejected since the fix is harmless, tested, and improves code consistency with no behavioral risk.

CI is green (the `claude-review` check, previously failing repo-wide per #535, passed on this run). No other outstanding feedback.

PR #587, opened 2026-09-02
fix(intl): add region-sensitive calendar preferences

### LINKED ISSUE #562: Intl.Locale getCalendars lacks region-sensitive preference data
## Summary

`Intl.Locale.prototype.getCalendars()` returns `["gregory"]` whenever no explicit `ca` keyword is present. ECMA-402 `CalendarsOfLocale` requires calendar preferences selected by language and `RegionPreference`, but jsse has no CLDR calendar-preference-by-region data yet.

This is a locale-data gap rather than the `rg`/`sd` resolution bug fixed by #551: wiring the shared region-preference helper into an unconditional `["gregory"]` result would not change observable behavior.

## Failing test262 coverage

- `intl402/Locale/prototype/getCalendars/likely-subtags-region.js`
- `intl402/Locale/prototype/getCalendars/region-override.js`
- `intl402/Locale/prototype/getCalendars/region-priority.js`
- `intl402/Locale/prototype/getCalendars/subdivision-region.js`

Eight scenarios fail (default and strict mode for each file).

## Expected implementation direction

Load or encode CLDR calendar preference data, then apply `RegionPreference` in the order required by ECMA-402. Representative data include Buddhist before Gregorian for Thailand and other region-specific preferences.

Split from #551 so that closing the region-override implementation does not lose this independent data gap.

pmatos:
## Summary

- add the CLDR territory calendar-preference ordering used by `CalendarsOfLocale`
- resolve locales through the existing `RegionPreference` signal order before selecting calendars
- preserve explicit `ca` keyword behavior and add literal region-data regression coverage

## Validation

- `./scripts/lint.sh`
- `cargo build --release`
- `cargo test --release` (619 passed, 1 ignored; integration tests passed)
- `uv run python scripts/run-test262.py -j 32` (99,903 / 99,907; 4 known unrelated failures, 0 regressions)
- `uv run python scripts/run-test262.py test262-extra/` (247 / 247)
- `uv run python scripts/run-custom-tests.py` (13 / 13)

Closes #562


pmatos:
Implementation note for reviewers: the stage plan's calendar snapshot omitted `XK`, but the current Unicode CLDR `calendarPreferenceData` places `XK` in the `gregory islamic-civil islamic-tbla` territory group, so this implementation includes it. The branch was also rebased over #572, which intentionally stopped tracking volatile test262 totals in README; the obsolete README-count commit was dropped to preserve that policy.

Issue #597, opened 2026-09-04
Parser: integer underflow panic (subtract with overflow) in parse_iteration_body on truncated for-loop source


pmatos:
Found by the new `cargo fuzz run parse_roundtrip` target while validating the fuzz tooling in #193 (a 20s smoke run against the freshly-committed seed corpus, no sanitizer, `-max_total_time=20`). Filing as a follow-up per the tooling PR's own triage rule: fuzzer findings are never fixed inside the tooling PR itself.

## Repro

Minimized by `cargo fuzz tmin` to 32 bytes:

```
for (v;7; i) {function t(a) {
 t
```

(Truncated/unterminated source — no closing braces, ends mid-token.)

```
$ cargo build   # debug profile has overflow-checks on; the plain release binary does not panic, since the overflow is silently wrapping there
$ ./target/debug/jsse repro.js
thread 'jsse-engine' panicked at src/parser/statements.rs:544:9:
attempt to subtract with overflow
```

Backtrace:
```
 3: <jsse::parser::Parser>::parse_iteration_body
             at ./src/parser/statements.rs:544:9
 4: <jsse::parser::Parser>::parse_for_statement
             at ./src/parser/statements.rs:1210:25
 5: <jsse::parser::Parser>::parse_statement_inner
             at ./src/parser/statements.rs:163:50
 6: <jsse::parser::Parser>::parse_statement
             at ./src/parser/statements.rs:99:22
 7: <jsse::parser::Parser>::parse_statement_or_declaration
             at ./src/parser/statements.rs:70:23
 8: <jsse::parser::Parser>::parse_program
             at ./src/parser/mod.rs:1061:29
```

## Analysis (not a fix — just what I saw while minimizing)

`parse_iteration_body` (src/parser/statements.rs:520-548) increments `self.in_iteration` before calling `self.parse_statement()` and unconditionally decrements it right after, before propagating any `Err` via `?`. Read in isolation that pairing is balanced per call frame, so the panic implies some other path mutates `in_iteration` asymmetrically when parsing is aborted partway through nested constructs (here: a `for` loop whose body is an unterminated `function` declaration) — worth tracing where `in_iteration` is touched elsewhere (e.g. function-body entry/exit resetting or restoring it) rather than assuming this call site itself is the culprit.

[...]

PR #598, opened 2026-09-04
feat(fuzz): add cargo-fuzz parser and differential-vs-node targets

### LINKED ISSUE #193: Tooling: add cargo-fuzz (parser + differential-vs-node targets)
Follow-up from the DevX audit in #192 — the single highest-value gap it identified. Split out of the CI-hardening PR because it needs a structural change, not just config.

**Blocker to resolve first:** the crate is currently binary-only (no `[lib]`/`src/lib.rs`), so there is nothing for libFuzzer to link against. A fuzz target needs a library target exposing the parser/interpreter entry points (e.g. `parse_program`).

Proposed work:
- [ ] Expose a `[lib]` target (add `src/lib.rs` re-exporting the needed modules, or a `[lib]` section) with `parser`/`interpreter` entry points reachable.
- [ ] `cargo fuzz init` + a `parse_roundtrip` target (feed arbitrary bytes → assert no panic/UB).
- [ ] A **differential** target comparing jsse vs `node` on the same source (surfaces divergences test262 misses — directly serves the 100% goal).
- [ ] Commit a seed corpus from `tests/` / `test262-extra/`.
- [ ] Short-budget `cargo fuzz run` smoke job in CI; optional nightly deep-fuzz.
- [ ] Package a fuzzing/differential `.claude` skill (mirroring the `performance-*` skills) so agents can drive runs and triage crashes.

pmatos:
## Summary

Adds cargo-fuzz to jsse: a `parse_roundtrip` target (parsing arbitrary
bytes must never panic) and a `differential` target (jsse vs `node` on
the same source), plus the structural change and CI/skill wiring both
need.

- Extracted `src/lib.rs` (issue #193's blocker): the crate was
  binary-only, so libFuzzer had no library target to link against.
  `main.rs` now consumes `ast`/`interpreter`/`lexer`/`parser`/`types`
  via `use jsse::{...}` instead of declaring its own `mod`s. No
  behavior change; widened three fields and added `is_empty()`/
  `Default` impls / `#[allow(clippy::should_implement_trait)]` where
  clippy's public-API lints started firing now that the crate is
  actually reachable from outside.
- Extracted `run_on_engine_stack` (the 128 MiB-stack thread spawn
  `main()` already did) into the lib, generic enough for a fuzz
  closure that borrows a `&str` per iteration with no `'static` bound.
- `parse_roundtrip`: a thin `fuzz_targets/parse_roundtrip.rs` calling
  `jsse::fuzz_parse_bytes`, which discards the parse `Result` (a parse
  error is a normal outcome for arbitrary bytes; `ast::Program` isn't
  `Send` anyway).
- `differential`: subprocess-based (not in-process — the interpreter
  has no step/instruction budget, so `while(1){}` would hang an
  in-process target). Runs `target/release/jsse` and `node` under
  `prlimit`, using the same `RLIMIT_AS` caps
  `scripts/run-test262.py` already uses per engine (512 MiB / 4 GiB —
  reusing jsse's own cap for node fatally OOMs it before it runs
  anything, verified empirically). Classifies results into three
  divergence tiers (Tier 1/2 panic, Tier 3 recorded) — see the ADR and
  the `Divergence Tier` entry in `CONTEXT.md`.
- Seed corpus copied from `tests/*.js` and `test262-extra/*.js` into
  `fuzz/corpus/`.
- `.github/workflows/fuzz.yml`: `fuzz-smoke` (PR gate) replays the
  corpus deterministically (`-runs=0`) only — see "Why fuzz-smoke is
  replay-only" below. `fuzz-deep` (workflow_dispatch + daily schedule)
  runs real new-mutation fuzzing for both targets and never blocks a
  PR. Validated locally with `actionlint` and `zizmor` (both clean);
  **this PR is fuzz-smoke's first real run** — I haven't seen it pass
  in actual GitHub Actions yet.
- `docs/adr/0004-fuzz-lib-target-and-subprocess-differential.md` (0004,
  not 0003 — see "Stale base" below) and `.claude/skills/fuzzing/SKILL.md`
  (run + triage workflow, mirroring `performance-analyse`/`performance-fix`).

## Issue #597: a real bug this tooling found, deliberately not fixed here

A 20-second local `-max_total_time=60`-style new-mutation smoke run
against `parse_roundtrip` — no sanitizer, the plain committed seed
corpus — found a real integer-underflow panic in the parser
(`parse_iteration_body`, on a truncated `for` loop). Filed as
[#597](https://github.com/pmatos/jsse/issues/597), not fixed in this
PR: this tooling's whole point is that a fuzzer finding is a lead for
its own issue/PR, fixed against the spec, never folded into the PR
that built the tooling and never fixed toward matching `node`'s
observed behavior.

**This is also why `fuzz-smoke`'s PR gate is corpus-replay-only
(`-runs=0`), not the "short bounded new-mutation pass" a first draft of
this plan had.** A mutation pass can hit #597 (or any other latent bug)
on any future unrelated PR, at random, which would make the fuzz gate
itself flaky. New-mutation fuzzing only runs in `fuzz-deep`
(scheduled/on-demand), where it can fail loudly without blocking
anyone's merge.

## Pre-existing test262 state (not introduced by this branch)

Full local `uv run python scripts/run-test262.py -j 32` run (with
`--baseline-ref` pointed at this branch's actual merge-base, see below)
shows:

- **22 `test262`/`intl402` "regressions"**, all
  `DateTimeFormat`/`Temporal.prototype.toLocaleString` locale-formatting
  tests (`dayPeriod`, `numbering-system`, `formatRange*`).
- **2 `test262-extra` failures**: `Eval-scoping.js:strict` and
  `with-gc-root.js:strict`.

Both reproduce **identically** on a clean checkout of this branch's
unmodified merge-base commit (`7a5389e`) in a scratch worktree — i.e.
they exist with zero code changes from this PR. The `intl402` cluster
looks environmental (icu4x locale data behavior differs from whatever
generated the `origin/main` baseline; `Cargo.lock` is unchanged by this
PR), not something this PR's diff could plausibly cause. Out of scope
for a fuzz-tooling issue; not touched here.

## Stale base

This branch's merge-base with `main` is `7a5389e`, and `main` has
advanced ~130 commits since (`perf-counters` clippy feature,
additional ADRs — hence `0004` here, not `0003`, since `0003` is
already taken on `main`). `git merge-tree --write-tree main HEAD`
reports **conflicts** in `src/interpreter/mod.rs`,
`src/interpreter/types.rs`, `src/lexer.rs`, `src/main.rs`, and
`src/types.rs` — expected, since both this branch and `main`
independently touched visibility/module structure in those files.
This PR will need a rebase onto `main` (with those five files resolved
by hand) before it can merge cleanly; flagging it here so it isn't
re-diagnosed from scratch.

## Also included

- `fix(intl)`: `needless_late_init` in `DurationFormat`'s sign-display
  branch — a drive-by fix needed to get `./scripts/lint.sh` green on
  this branch's base commit before touching anything fuzz-related
  (the toolchain here is newer than whatever last verified `lint.sh`
  on this exact base; unrelated to `DurationFormat` behavior).

## Known limitation (non-blocking, for the record)

`run_with_timeout` (`fuzz/src/lib.rs`) reads a child's piped stderr
only after it exits. A child that emits more than the OS pipe buffer
(~64 KiB) before hanging would block on the pipe rather than being
killed at the wall-clock deadline, and would classify as `TimedOut` →
`Recorded`. This is a false-negative only (a missed finding), never a
spurious Tier 1/2, and no seed in the committed corpus is anywhere
near that output volume.

## Test plan

- [x] `./scripts/lint.sh`
- [x] `cargo build --release`
- [x] `cargo test --release` (291 lib tests, was 0 lib / N bin before
      the `src/lib.rs` split)
- [x] `fuzz/`: `cargo fmt --check`, `cargo clippy --all-targets -- -D
      warnings`, `cargo test --lib` (13 tests) — all clean
- [x] `cargo +nightly fuzz build parse_roundtrip|differential
      --sanitizer none`
- [x] `cargo +nightly fuzz run parse_roundtrip -- -runs=0
      fuzz/corpus/parse_roundtrip` (34/34 seeds clean)
- [x] `cargo +nightly fuzz run differential -- -runs=0
      fuzz/corpus/differential` (7/7 seeds clean; each seed manually
      verified against both engines with `--expose-gc`/prelude first)
- [x] Full `run-test262.py -j 32 --baseline-ref 7a5389e`: 99,546/99,568
      (99.98%), 323 new passes, 22 pre-existing regressions (see above)
- [x] `run-test262.py test262-extra/`: 53/55 (2 pre-existing failures,
      see above)
- [x] `run-custom-tests.py`: 6/6
- [x] `actionlint` and `zizmor` against `.github/workflows/fuzz.yml`
      (clean)

Closes #193


pmatos:
Autofix paused: total CI budget exhausted (`TOTAL_CI_BUDGET` = 15 min).

**Fixes landed this run:**
- `b6a577f` — merged `main` into this branch (6 real conflicts: `CONTEXT.md`, `src/interpreter/mod.rs`, `src/interpreter/types.rs`, `src/lexer.rs`, `src/main.rs`, `src/types.rs`), plus made the PR's own pre-existing `src/lib.rs` extraction (commit f09f9ac) actually compile and pass tests: `main.rs`'s CLI logic moved into a new `src/cli.rs` module inside the library crate (keeping `Interpreter`/`Parser`/`Completion`/the AST at their original `pub(crate)` visibility, avoiding a much larger public-API expansion), removed two dead `is_empty()` methods added only to satisfy a clippy lint on now-unreachable types, and gated two new unit tests that recurse 5000-deep into nested array literals behind `#[cfg_attr(debug_assertions, ignore = "..."])` — they SIGABRT via native stack overflow in **debug** builds only (`cargo test`'s profile), well below the parser's own catchable `MAX_PARSE_DEPTH` guard; both pass cleanly under `--release`. Filed #599 to track that debug-mode gap separately.
- `8d48862` — fixed a regression caught by this run's own dual-evaluator regression review: the merge's `git add -A` staged the `test262` submodule gitlink from a stale working-tree checkout instead of git's correctly auto-merged index value, silently reverting main's PR #547 submodule bump. Restored.

**Feedback resolved:** none pending — no review threads, review summaries, or actionable PR comments existed on this PR (the one existing PR comment is Codex's own "usage limit reached" notice from its automated review integration, not reviewer feedback).

**Still pending:** GitHub has not yet registered the `8d48862` push as this PR's head — `gh api repos/pmatos/jsse/pulls/598` and the PR's own `/commits` listing still report `b6a577f` as the tip nearly 10 minutes after `8d48862` was pushed (confirmed present on the remote ref via `git ls-remote` and `gh api .../git/refs/...`), so no fresh CI run has been triggered for it and mergeability is stuck at `unknown`/`null`. This looks like a GitHub-side processing stall unrelated to the code; `8d48862` only touches the `test262` submodule pointer and is a self-evidently correct one-line fix, independently verified clean by a fresh regression-review pass. CI was fully green (build, clippy, fmt, tests, fuzz-smoke, CodeQL, cargo-deny, Semgrep, Claude Code Review) on the prior head `b6a577f`.

Resume with `/pm-autofix-pr 598`.

PR #602, opened 2026-09-05
fix(parser): restore context counters when a body parse fails

### LINKED ISSUE #597: Parser: integer underflow panic (subtract with overflow) in parse_iteration_body on truncated for-loop source

pmatos:
## Summary

- A function body, class static block and `switch` case block each re-scope the parser's `in_iteration` / `in_switch` / `in_non_arrow_function` counters, but only restored the saved values on the success path. Truncated source aborts the parse mid-body, so the zeroed counters escaped into the enclosing construct and a decrement ran that its own frame never matched — `attempt to subtract with overflow` under `overflow-checks`, silent wrapping otherwise.
- Each of the three sites now runs its fallible middle through a helper, captures the `Result`, restores the saved context unconditionally, and only then propagates the error.
- The minimized fuzz input plus a static-block and a `switch` variant are seeded into `fuzz/corpus/parse_roundtrip/`, which the `fuzz-smoke` PR gate replays with `-runs=0` under `overflow-checks`.

### Two underflow sites, not one

The issue reports the `in_iteration` panic in `parse_iteration_body`. Code review turned up a second, distinct one that the same fix closes:

```
export default function () { class C { static {
```

`parse_function_body_with_context` (`declarations.rs:1243`, reached only from `modules.rs:411`/`:439`) decrements `in_non_arrow_function` **unconditionally**, and the static-block arm zeroes that counter — so pre-fix this panicked at `declarations.rs:1256`, on a path `parse_program` cannot reach at all. Both paths are covered by the regression test.

### Scope notes

- **`in_non_arrow_function` at the other eight sites** is still left elevated by an aborted parse. Those decrements all sit after a `?`, so they are skipped rather than unmatched — untidy, not an underflow. Not fixed here.
- **The `switch` refactor is hygiene, not a fix.** `in_switch` could not underflow before: its only decrement sat after the case-loop `?`. Making it unconditional deliberately trades a benign high-leak for a loud underflow, matching `parse_iteration_body`, and is what makes the new `in_switch == 0` assertion meaningful. Recorded in a comment at the site.
- **The duplication that allowed this bug** — two sites hand-saving ~12 context fields, four more repeating a smaller version — is real and is filed as a follow-up rather than expanded into this PR.
- **The issue's "the parser may already be misbehaving silently on release builds" concern does not hold.** The leak only occurs on a path that aborts the whole parse, and the `Parser` is then discarded; the three backtracking sites only peek tokens and never enter a body. No release-build ECMAScript behaviour changes — which is why the regression test asserts the counters directly in `src/parser/mod.rs` (profile-independent) rather than living in `test262-extra/`.

Fixes #597

## Test plan

- [x] Issue's 32-byte repro on a debug build — `SyntaxError`, no panic (pre-fix: panic at `statements.rs:544`)
- [x] Module repro above on a debug build — `SyntaxError`, no panic (pre-fix: panic at `declarations.rs:1256`)
- [x] All three fix sites verified load-bearing by reverting each file in turn and re-running the new test
- [x] `cargo test --release` — 646 unit + integration tests, 0 failures
- [x] `./scripts/lint.sh` — rustfmt + clippy (default and `perf-counters`) clean
- [x] `uv run python scripts/run-test262.py -j 32` — 99,568/99,568 (100.00%), 0 regressions
- [x] `uv run python scripts/run-test262.py test262-extra/ -j 32` — 308/308


pmatos:
## Code review — two axes (`/code-review xhigh`)

Reviewed `origin/main...HEAD` (`a0ef691`) on two independent axes in parallel. Both reports below; findings are not merged or reranked across axes, deliberately.

## Standards

**(a) Documented-standard violations — none.** Two rules that could have bitten, both checked and clear:

- **Fuzz corpus** — `docs/adr/0004-*.md` deliberately declines a "`fuzz/corpus/<target>/`-is-pristine guarantee", and the fuzzing skill says *"if a specific discovered input is worth keeping as a regression seed (e.g. after fixing the bug it found), add it deliberately"*. The three seeds are exactly that case. Conformant.
- **Test placement** — `CLAUDE.md`'s `test262-extra/` rule is scoped to *"when the failure changes an observable ECMAScript value or throw"*. Not triggered: no `[profile]` in `Cargo.toml`, so release has `overflow-checks` off, and the parser has no statement-level backtracking (only single-token peeks). `parse_program` returns `Err` before and after. A unit test + fuzz seed is the only possible home.

**(b) Baseline smells (judgement calls)**

1. **Data Clumps / Duplicated Code** — the two big sites each hand-save/zero/restore the same ~12-field context; the diff's own comment names the duplication. Filed as #608 rather than expanded here.
2. **Comment inaccuracy** — *"every one of its decrements sits after a `?`"* is false at `declarations.rs:1243`. **Fixed** in `11732e3`.
3. **Minor** — `cases.push(SwitchCase { test, consequent: consequent? })` buried `?` in a field initialiser. **Fixed** in `11732e3`.
4. **Asymmetry** — `parse_function_body_inner` pulls `eat(RightBrace)` inside the guarded region via `.and_then`; the static-block site leaves it outside. Verified behaviour-preserving; unifying it needs #608's guard.

**Explicitly not smells:** the extracted helpers are *not* Middle Man — each carries real §14.12.1/§15.2.1 validation. `parse_switch_case_block`'s capture-restore-propagate replicates the pre-existing house idiom at `statements.rs:542-545` verbatim.

## Spec

**Verification of the three claims in the PR body:**

1. **Every error path restores — HOLDS.** In all three sites the only `?` before the re-scoping precedes any counter save; the fallible middle is captured and only returned after every restore. No leaking `Err` path found.
2. **"Every `in_non_arrow_function` decrement sits after a `?`" — REFUTED.** 8 of 9 do; `parse_function_body_with_context` (`declarations.rs:1242-1245`) decrements unconditionally, reachable from `modules.rs:411`/`:439`. Safety there now depends on this PR's own static-block fix, not on the stated invariant. **The claim was wrong in both the PR body and the test comment; both corrected in `11732e3`, and the module path is now covered by the test.**
3. **Reordering is behaviour-preserving — HOLDS.** The only parser→lexer coupling is `lexer.strict` and `lexer.is_module`; `set_strict(prev_strict)` still runs *after* `eat(RightBrace)`, so the token following `}` is lexed under identical lexer state. None of the 12 flags that moved after `eat` are read by the lexer. The new error-path `set_strict` is unobservable — no caller swallows a body-parse `Err`.

**(a) Partial:** four sibling sites with the identical shape are untouched — `parse_block_statement` (`statements.rs:231`/`:262`), `parse_try_statement` (`:1302`/`:1310`), catch (`:1338`/`:1346`), finally (`:1375`/`:1383`) — each leaking `in_block_or_function`. Harmless today (parse aborts), but the diff's own rationale applies equally. Tracked in #608.

**(b) Scope creep:** the `switch` refactor is not required for #597 — `in_switch` could not underflow pre-fix, and the change *introduces* an unconditional decrement where none existed, converting a future benign leak into an underflow. Kept deliberately (it matches `parse_iteration_body` and makes the new assertion meaningful); the trade-off is now recorded in a comment at the site and in the PR body.

---

**Standards: 0 hard violations, 4 judgement calls. Spec: 1 refuted claim (the worst finding on that axis), 1 partial, 1 scope note.** Fixes for the two correctable findings and the refuted claim are in `11732e3`; the structural finding both axes converged on is #608.

pmatos:
## `/simplify` — 4 cleanup agents

Quality pass (reuse / simplification / efficiency / altitude) over `origin/main...HEAD`. Applied in `478b61b`; test262 re-run after, still 99,568/99,568.

**Applied**

1. **Altitude** — the switch comment claimed the unconditional decrement "underflows loudly", which only holds under `overflow-checks`; in release it wraps. Corrected.
2. **Altitude** — the invariant that made this bug possible ("only a counter whose decrement runs unconditionally somewhere turns a leaked zero into an underflow") was written nowhere, so the next counter added reintroduces it. Now recorded at the counter declarations in `mod.rs`.
3. **Simplification** — the capture-restore-propagate idiom was spelled two ways (`let stmts = result?` vs shadowing `let cases = cases?`). Unified on the `result`-named form at all three sites.
4. **Simplification** — `.and_then(|body| self.eat(..).map(|()| body))` → the plainer `?`/`Ok(body)` block.
5. **Simplification** — shared the three repeated assertions in the regression test.

**Skipped, with reason**

- **Delete `parse_switch_case_consequent`** and hoist `in_switch_case` up to `parse_switch_statement` (−18 lines). The agent flagged it "wants a test run" — it widens `in_switch_case = true` over `case <expr>:` test parsing. It is very likely inert (its only reads are the Annex B statement-position checks at `statements.rs:8,17`, and expressions can't reach them), but semantically widening a flag's scope inside a PR specifically about flag scoping is the wrong trade for 18 lines.
- **Four copies of the VarDeclaredNames/LexicallyDeclaredNames check**, and `parse_static_block_statements` duplicating `parse_block_statement`'s body. Both real, both **pre-existing** — the extraction surfaced them rather than creating them — and a proper fix touches `mod.rs:936` and `statements.rs:247`, outside this diff, while changing error construction at two sites. Moved to #608.
- **Normalizing the other 8 `in_non_arrow_function` sites** — that is the deferred refactor. #608.

**Clean on their angle**

- **Efficiency: no findings.** The three extractions force no new allocation, `clone()`, or lost in-place mutation; `saved_param_names.as_ref()` shortens a borrow without changing the drop point; the error-path restores only pull forward a dealloc `Parser`'s own drop would do, and the error path is terminal (the only error-swallowing lookaheads are token-level and never enter these functions). Three fuzz seeds at 32/29/20 bytes, one per fix site, no overlap.
- **Reuse: the new test is clean** — the existing `parse()` helper unwraps and discards the `Parser`, so it can't serve a test that asserts on counters afterwards; the inline form matches the existing error-test idiom.
- **Altitude: the fix is at the right depth.** Restore-on-error is the root cause, not a symptom patch; threading the counters as parameters would touch nearly every `parse_*` signature across five files. Three-of-seven is the *right* cut — the other four re-scope only `bool`s, which cannot underflow.

**One correction the pass produced for #608 itself:** its original `SavedContext` snapshot proposal is a notch too shallow — it dedups the fields but leaves "forgot to restore on the error path" reachable. Retargeted to a `scoped()` combinator that owns the restore point, which removes the bug class structurally and gets the field dedup for free.

Issue #331, opened 2026-07-20
GC rooting ergonomics: RAII frame guard, single mutation boundary, and split frame-roots vs registered-roots


pmatos:
## Motivation

Reconciling #290 (the redundant `eval_array_literal` / member-assignment rooting idiom) surfaced a bigger, longer-lived question than inline-vs-helper style: the GC temp-root machinery relies on *conventions* that the compiler does not enforce, and `gc_temp_roots` currently mixes two different ownership models. This issue tracks the durable design work; #290 handled only the one-line idiom reconciliation.

The `gc_root_value` helper is now the intended mutation boundary for rooting a single `JsValue`. The items below build on that.

## 1. The real risk: manual frame lifetime is a convention, not an invariant

The canonical rooting pattern is:

```rust
let gc_frame = self.gc_root_frame();
let result = (|| {
    // ... eval that may hit GC safepoints; abrupt returns live inside the closure
})();
self.gc_unroot_frame(gc_frame);
result
```

This works, but it is a discipline the programmer must reproduce correctly every time. A future edit can:

- add an outer-function `return` that bypasses `gc_unroot_frame`,
- forget cleanup on a newly added branch,
- pass the wrong frame marker, or
- truncate an enclosing frame.

The failure modes are either **premature collection** (a root dropped too early) or a **root leak** (roots never truncated → masks bugs, grows memory). Both are subtle and hard to catch by reading.

### Proposed: RAII frame guard

Encapsulate the frame so cleanup runs on every exit path (including panics) via `Drop`:

```rust
struct GcRootFrame<'a> {
    interp: &'a mut Interpreter,
    depth: usize,
}

impl Drop for GcRootFrame<'_> {
    fn drop(&mut self) {
        self.interp.gc_temp_roots.truncate(self.depth);
    }
}
```

Implement `Deref<Target = Interpreter>` + `DerefMut` and a focused `root_value`, so callers use the interpreter through the guard:

```rust
let mut frame = self.gc_root_scope();
frame.root_value(&obj_val);
let rval = frame.eval_expr(right, env);
// every return / panic drops `frame`, which unroots.
```

Advantages: early returns auto-unroot; nested frames follow Rust drop ordering; no `unsafe`; the guard can expose only supported rooting ops; manual `gc_unroot_frame` calls disappear from migrated code. Cost: code inside the scope must call through `frame` rather than `self`.

[...]

PR #604, opened 2026-09-05
fix(iterator): keep concat current iterator GC-rooted

### LINKED ISSUE #584: bug(iterators): concat current-iterator state is not GC-rooted after reassignment
## Summary

`Iterator.concat` has the same native-closure GC-rooting bug identified for `Iterator.prototype.flatMap` in #556.

Its helper state stores `current_iter` and `current_next` in `Rc<RefCell<...>>` fields that start as `None`. `set_helper_gc_roots` runs at helper construction while those fields are still empty. Later, `next_fn` assigns a newly opened iterator and next method into the closure-only state, but neither value is reachable by the GC tracer.

Current reassignment site:

```rust
state_next.borrow_mut().2 = Some(new_iter);
state_next.borrow_mut().3 = Some(next_method);
```

The construction-time `if let Some(...)` branches for fields `.2` and `.3` are therefore dead code.

## Impact

A collection between helper `.next()` calls can reclaim the current iterator when it is otherwise ephemeral. The next call then uses a stale object id. This was confirmed as a real failure mode for the equivalent `flatMap` state in #556.

## Suggested direction

Follow #556's bounded-root pattern: create one traced two-slot container, root that container once on the iterator helper, and mutate its slots in place whenever the current iterator changes. Do not call `pin_native_root` on every reassignment; pins only accumulate and would retain every iterator opened over the helper's lifetime.

Add a `$262.gc()` regression under `test262-extra/` that opens an ephemeral iterable, consumes one value, collects, and resumes the same current iterator.

Found while implementing #556; deliberately kept out of that single-issue branch.

pmatos:
## Summary

- hold `Iterator.concat`'s current underlying iterator and its `next` method in one traced two-slot container (`RootedPair`) rooted on the helper, instead of in the native closure's own `Rc<RefCell<...>>` state that the collector never walks
- mutate and clear that fixed container in place, so a replacement stays visible to the collector without accumulating a permanent pin per iterable opened
- factor the container into a newtype shared with `flatMap` (fixed in #585), so the pattern has one implementation rather than two copies of the same inline block

Before this change, a collection between two `next()` calls could reclaim an otherwise ephemeral current iterator and leave the helper resuming a stale object id:

```js
var it = Iterator.concat([1, 2, 3], [4, 5]);
it.next();     // { value: 1, done: false }
$262.gc();
it.next();     // TypeError: Iterator result is not an object
```

Fixes #584

## Also in this PR: a flatMap behaviour change

Keying the container's occupancy on slot 0 alone (the iterator) rather than on both slots also corrects `Iterator.prototype.flatMap`. Since #585 it has *silently skipped* an inner iterator whose `next` method is not callable; per `GetIteratorFlattenable` -> `GetIteratorDirect` there is no callability check, so the *TypeError* is owed by the first `IteratorStepValue`. Node agrees:

```js
[1, 2].values().flatMap(x => ({ [Symbol.iterator]: () => ({}) })).next();
// origin/main: { done: true }      node & this PR: TypeError: undefined is not a function
```

It is the same line of the shared helper, so it is corrected here rather than deferred — but it is outside #584's stated scope, hence this note.

Full review notes, including the findings applied and declined, are in the [review comment](https://github.com/pmatos/jsse/pull/604#issuecomment-5554904603).

## Test plan

Four regression tests under `test262-extra/`:

- `iterator-concat-current-iterator-gc-rooting.js` — resumes the current iterator, and opens the next iterable, across collections (fails on `origin/main`: `TypeError: Generator.prototype.next called on non-object`)
- `iterator-concat-return-closes-current-iterator-after-gc.js` — `return()` still closes the current iterator after a collection (fails on `origin/main`: the iterator is never closed)
- `iterator-flatMap-non-callable-next-method.js` — the absent `next` throws instead of ending the helper (fails on `origin/main`: `{done:true}`)
- `iterator-concat-non-callable-next-method.js` — the absent `next` throws and `[Symbol.iterator]` runs exactly once. **Passes on `origin/main`**; it guards a regression introduced and fixed during review, which never shipped.

- [x] `./scripts/lint.sh` — clean
- [x] `cargo build --release`
- [x] `cargo test --release` — 645 unit + integration tests, 0 failed
- [x] `uv run python scripts/run-test262.py --fail-on-failures test262/test/built-ins/Iterator/ -j 32` — 1,028/1,028
- [x] `uv run python scripts/run-test262.py --fail-on-failures test262-extra/ -j 32` — 316/316
- [x] `uv run python scripts/run-test262.py -j 32` — 99,568/99,568, 0 regressions against `origin/main`

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -53,3 +53,11 @@ _Avoid_: root scope marker, gc stack pointer.
 **GC Root Scope**:
 The lexical scoping of a Temp-Root Frame behind the `with_gc_root_scope(|interp| …)` combinator: it captures the frame, runs the body, and truncates on every exit path (tail, early `return`, `?`) so the teardown cannot be forgotten on a branch. Prefer it to a hand-paired `gc_root_frame`/`gc_unroot_frame` for a whole-body, single-frame native; reach for the raw primitive only when frames nest or interleave across early exits.
 _Avoid_: root guard, unroot epilogue.
+
+**Pinned Native Root**:
+A `JsValue` attached to an anchor object's `gc_native_roots` list by `pin_native_root`, so it stays reachable for as long as the anchor is. Unlike a Temp-Root Frame it outlives the native call that created it, which is what a native closure's captures need. Pins only ever accumulate, so an anchor must be pinned to a *fixed* set of values, established once.
+_Avoid_: permanent root, closure root.
+
+**Rooted Slot**:
+A GC-traced container an anchor pins once and the owner then mutates in place, for a capture whose value is *replaced* over the anchor's lifetime. A native closure's own `Rc<RefCell<…>>` state is invisible to the tracer, and re-pinning each replacement would retain every superseded value, so the slot — not the value — is what gets pinned. `RootedPair` in `builtins/iterators.rs` is the two-slot case: the iterator an iterator helper is currently drawing from, plus that iterator's `next` method.
+_Avoid_: root cell, traced box, rooted buffer.
--- a/src/interpreter/builtins/iterators.rs
+++ b/src/interpreter/builtins/iterators.rs
@@ -341,6 +341,48 @@ fn iterator_close_getter(interp: &mut Interpreter, iterator: &JsValue) -> Result
     }
 }
 
+// A two-slot container an iterator helper roots once and mutates in place.
+// Native-closure captures that are replaced after the helper is built cannot be
+// stored in the closure's own `Rc<RefCell<...>>` state: the collector never
+// walks it, so a replacement is unreachable and can be collected mid-iteration.
+// Pinning each replacement instead would retain every superseded value for the
+// lifetime of the helper. See `Interpreter::set_helper_gc_roots`.
+fn rooted_pair_new(interp: &mut Interpreter) -> JsValue {
+    interp.create_array(vec![JsValue::UNDEFINED, JsValue::UNDEFINED])
+}
+
+fn rooted_pair_get(interp: &Interpreter, pair: &JsValue) -> (Option<JsValue>, Option<JsValue>) {
+    let pair_id = pair
+        .as_object_id()
+        .expect("rooted pair must be an Array object");
+    let cell = interp.get_object_cell_expect(pair_id);
+    let obj = cell.borrow();
+    let elements = obj
+        .array_elements()
+        .expect("rooted pair must have Array elements");
+    (
+        elements.first().filter(|v| !v.is_undefined()).cloned(),
+        elements.get(1).filter(|v| !v.is_undefined()).cloned(),
+    )
+}
+
+fn rooted_pair_set(interp: &Interpreter, pair: &JsValue, first: JsValue, second: JsValue) {
+    let pair_id = pair
+        .as_object_id()
+        .expect("rooted pair must be an Array object");
+    let cell = interp.get_object_cell_expect(pair_id);
+    let mut obj = cell.borrow_mut();
+    let elements = obj
+        .array_elements_mut()
+        .expect("rooted pair must have Array elements");
+    elements[0] = first;
+    elements[1] = second;
+}
+
+fn rooted_pair_clear(interp: &Interpreter, pair: &JsValue) {
+    rooted_pair_set(interp, pair, JsValue::UNDEFINED, JsValue::UNDEFINED);
+}
+
 fn close_iterator_for_error(
     interp: &mut Interpreter,
     iterator: &JsValue,
@@ -2889,10 +2931,7 @@ impl Interpreter {
                     Err(e) => return Completion::Throw(e),
                 };
 
-                let inner_roots = interp.create_array(vec![
-                    JsValue::UNDEFINED,
-                    JsValue::UNDEFINED,
-                ]);
+                let inner_roots = rooted_pair_new(interp);
                 // state: (outer_iter, outer_next, mapper, counter, rooted inner pair, alive, running)
                 #[allow(clippy::type_complexity)]
                 let state: Rc<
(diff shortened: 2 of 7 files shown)


pmatos:
## Two-axis code review (`origin/main...HEAD`, xhigh)

Standards and Spec axes reviewed in parallel. One blocker found and fixed in `aa154b4`; five of seven Standards findings applied, two declined with reasons.

### Blocker (Spec axis) — fixed

The two-slot container keyed occupancy on *both* slots being non-`undefined`, conflating "empty slot" with "`next` is `undefined`". Per `sec-iterator.concat` step 3.a.iii, `GetIteratorDirect` performs no callability check — an opened iterator with no `next` method is a valid Iterator Record, and the *TypeError* is owed later by `IteratorStepValue`. Because `concat` does not advance `idx` when it opens an iterable, reading the slot as empty sent control back to "open the next iterable" at the same index, re-invoking `[Symbol.iterator]` forever.

```js
Iterator.concat({ [Symbol.iterator]() { return {}; } }).next();
```

| | `origin/main` | first cut (`61f3e8e`) | now (`aa154b4`) |
|---|---|---|---|
| `Iterator.concat` | TypeError, opened x1 | **infinite loop** | TypeError, opened x1 |
| `Iterator.prototype.flatMap` | `{done:true}` | `{done:true}` | TypeError |

Occupancy now derives from slot 0 alone; slot 1 is stored and returned verbatim.

**This also changes `Iterator.prototype.flatMap`, which is a behaviour change to code #585 shipped.** Since #585, flatMap has *silently skipped* an inner iterator whose `next` is not callable instead of throwing. Node throws `TypeError: undefined is not a function` and the spec agrees (`GetIteratorFlattenable` -> `GetIteratorDirect`, no callability check). It is the same line of the shared helper, so it is corrected here rather than deferred; flagging it explicitly because it is outside #584's stated scope.

### Standards findings applied

- **`info:` blocks now quote the algorithm** and `features:` corrected to `[iterator-sequencing, host-gc-required]` (all 32 upstream `Iterator/concat` tests use `iterator-sequencing`; `iterator-helpers` was wrong). CLAUDE.md requires extra tests to "include spec part that is tested and follow the exact same patterns of test262 tests."
- **`CONTEXT.md` glossary gap closed.** Per `docs/agents/domain.md`, a concept absent from the glossary is either invented language or a real gap. This is a real gap: `Pinned Native Root` and `Rooted Slot` are added to the Memory section beside `Temp-Root Frame` / `GC Root Scope`, which cannot express a capture that outlives a single native call.
- **`struct RootedPair`** replaces the four free functions (Feature Envy / Primitive Obsession), establishing the "two-element Array" invariant once instead of re-`expect()`ing it at each access. This also removes the get/set contract split — `get` tolerated a short array via `.first()`/`.get(1)` while `set` indexed and panicked.
- Container comment no longer restates the doc-comment `8ffe5a3` added to `set_helper_gc_roots`.

### Declined

- **Generalising the accessor to `chunks`/`windows`** (10 further inline `.expect()` sites): those buffers are variable-length accumulators (push/clear/clone/len), a different shape from a fixed two-slot pair. Sharing only the two-line preamble would extend this PR into two unrelated helpers.
- **Naming the `Rc<RefCell<tuple>>` state**: the positional 5-tuple is the file's pre-existing house shape across every helper; renaming it is a file-wide refactor, not this change.

### New tests

- `iterator-concat-non-callable-next-method.js` — asserts the TypeError and that `[Symbol.iterator]` runs exactly once. **Passes on `origin/main`**: it guards the regression above, which never shipped.
- `iterator-flatMap-non-callable-next-method.js` — **fails on `origin/main`** (returns `{done:true}`), so it covers a real shipped bug.

### Validation after the fixes

- `./scripts/lint.sh` — clean
- `cargo test --release` — 645 unit + integration, 0 failed
- `test262/test/built-ins/Iterator/` — 1,028/1,028
- `test262-extra/` — 316/316
- `run-custom-tests.py` — 11/11
- full `run-test262.py -j 32` — 99,568/99,568, 0 regressions

### Clean on both axes

flatMap's conversion is faithful to `8ffe5a3` at all five call sites; the clear-after-`iterator_close_getter` reordering in `return` matches the precedent and stays guarded by `running`/`alive`; root vector capacity is correct; no other reassigned-capture sites remain in `iterators.rs` (`zip`/`zipKeyed` open everything eagerly and root at construction).

pmatos:
## `/simplify` pass — 4 parallel cleanup agents (reuse / simplification / efficiency / altitude)

Applied in `76f279d`. Quality only — no behaviour change; the fix and its tests are unchanged. **Net -55 lines** against the previous head.

### Applied

**The rooted pair does not belong in the state tuple.** Simplification and Efficiency reached this independently. `RootedPair` is created once and never reassigned — I checked every `borrow_mut().N =` in both closures — so tuple membership bought nothing and cost a clone out of the `RefCell` every loop iteration. It is now captured per closure the way the state handle already is. concat's tuple drops 5 -> 4 fields and flatMap's 7 -> 6, which **undoes part of the renumbering `aa154b4` performed**.

**concat's per-iteration destructure collapses to `let idx = state_next.borrow().1;`.** Two things fell out with the pair:
- the in-loop `if !alive` re-check was unreachable — every site that clears the flag `return`s immediately (verified across all six).
- `s.0.clone()` cloned the entire `Vec<(JsValue, JsValue)>` of every iterable on every loop iteration, to read `len()` and one entry. Now `state_next.borrow().0.get(idx).cloned()`. For `Iterator.concat(a, b)` over 100k elements that was ~100k allocations plus a Drop pass over 200k values. **This predates the branch** (verbatim on `origin/main`) — included only because it lives in the tuple this branch rewrites.

**Smaller:** `as_root() -> &JsValue` became `root() -> JsValue` (both callers cloned immediately); the doc comment dropped the paragraph restating `CONTEXT.md` and gained the reason the slot is arena-backed rather than a Rust container — `ObjectHandle::borrow_mut` runs `remember_if_old()` (`object_arena.rs:73-75`), so an old helper taking on a young iterator enters the remembered set, which a plain container would silently skip; and the four test `info:` blocks are trimmed to the steps each test actually exercises (one had transcribed 20 lines of algorithm to support one sentence).

**Altitude cross-reference.** `pin_native_root`'s doc now names the replaced-capture case and points at `RootedPair`. `promise.rs` had already paid for that gap, hand-rolling a weaker `value_anchor` workaround at five sites.

### Altitude verdict: depth is right, no rewrite

The tempting deeper fixes — a `Vec<JsValue>` slot on `IterHelperData`, or a traced handle the closure captures directly — are both wrong here: any container mutated outside the arena bypasses the generational write barrier, and an old anchor gaining a young iterator would not be remembered. Two consumers do not justify new engine machinery.

### Skipped, with reasons

- **`with_array_elements[_mut]` across the 9 borrow-preamble sites** — real duplication, but 6 are in `chunks`/`windows`, outside this diff; an `Interpreter` API used by only the 3 new sites leaves the pattern half-converted. Filed as #610.
- **Move `RootedPair` beside `set_helper_gc_roots`** — not possible: that is at `iterators.rs:1510` inside `impl Interpreter`, and a module-level struct cannot sit between impl methods. Handled with cross-references in both directions.
- **Double arena lookups (`get`+`clear`, `get`+`set`) and `create_array`'s three unread property-map entries** — both sub-1% against the JS call they bracket; Efficiency judged the Array the defensible container.
- **Promote to `gc.rs` as `RootedSlots` + migrate `promise.rs`** — real work, not this PR. Filed as #609, noting it is the registered-roots half of #331.

### Re-validated after the refactor

`./scripts/lint.sh` clean; `cargo test --release` 645 passed; Iterator 1,028/1,028; `test262-extra/` 316/316; custom 11/11; full test262 **99,568/99,568, 0 regressions**. The four behaviour probes (GC repro, concat and flatMap non-callable `next`, open-count) all unchanged.

Issue #361, opened 2026-07-21
tweetnacl-js Node-compat harness: exhaustive vector coverage (currently sampled)


pmatos:
Follow-up from #300 (tweetnacl-js Node-compat harness).

## Context

tweetnacl-js's curve25519/Ed25519 point arithmetic is extremely expensive on
jsse's tree-walking interpreter relative to V8 — roughly 100-3500x slower per
operation (measured: a single `scalarMult.base` call ≈3.4s on jsse vs ≈35ms on
Node; `sign.detached.verify` ≈12s vs ≈76ms). At the full upstream vector counts
(256 scalarmult vectors, 256 box vectors, 1024 Ed25519 sign vectors) the full
`test-node` suite is estimated at roughly 7 hours wall-clock, which isn't
practical to actually run as part of landing the harness.

A correctness smoke test (all 13 upstream test files, vector arrays truncated
to 5 entries each) passed 1233/1233 assertions on jsse, byte-identical to
Node — so this is confirmed to be pure interpretation overhead, not an engine
correctness bug.

## What #300 ships

The harness (`scripts/libs/tweetnacl-js.sh`) evenly samples the three
curve-heavy vector files (`scalarmult.random.js`, `box.random.js`,
`sign.spec.js`) down to 20 vectors each (stride-sampled across the full
array, not just a prefix). Every other data file (secretbox, hash,
onetimeauth — no elliptic-curve cost) keeps its full upstream vector count.
This is the first sampled corpus in the Node-compat harness — every other
library config runs its suite unmodified.

## Ask

Once the engine has a meaningfully faster numeric path for this kind of
code (e.g. further bytecode-VM work under #67, or a dedicated fast path for
typed-array-heavy field arithmetic), revisit `scripts/libs/tweetnacl-js.sh`
and either raise the sample sizes or run the full upstream vector counts,
removing the sampling note from `scripts/README.md`.

pmatos:
**Decision (via /pm-triage, 2026-09-04):** Benchmark curve25519/Ed25519 point arithmetic under the bytecode VM (issue #67, merged shortly before this issue was filed) — if it meaningfully closes the perf gap, raise/remove the sample caps in `scripts/libs/tweetnacl-js.sh` and drop the sampling note from `scripts/README.md`. If the bytecode path doesn't help enough for this typed-array-heavy field-arithmetic workload, leave the sampled harness in place as a long-term, perf-gated deferral rather than pursuing a dedicated fast path now.

[...]

Issue #603, opened 2026-09-05
bytecode: compile `new` expressions and compound member assignment (blocks typed-array field arithmetic)


pmatos:
Split out of #361, which measured the bytecode VM against tweetnacl-js's
curve25519/Ed25519 field arithmetic and found it makes no difference
(1.00-1.07x, inside run-to-run noise on a shared host).

The `perf-counters` build shows the reason is not that the VM is slow on this
code — it is that the VM never runs it. Three functions carry **96% of the
tree-walker's work** in a `scalarMult.base` run, and all three bail out of the
compiler. Their work-unit counts are *identical* with and without
`--bytecode` — the direct evidence the VM never reaches them — so they end up
holding 99% of everything still running on the tree-walker:

| minified | tweetnacl name | share of AST work left under `--bytecode` | bail reason |
|---|---|---|---|
| `X` | `M` — the GF(2^255-19) multiply | 74.07% | `expression:New` |
| `C` | `car25519` — carry propagation | 22.88% | `assign target` |
| `F` | `sel25519` — constant-time swap | 2.06% | `assign target` |

With `--bytecode` the VM executes 5,584,177 ops while 136,962,403 AST work
units remain — it displaces **3.3%** of the work despite compiling 29% of
function *invocations*. The same pattern holds on the non-curve control
(SHA-512): 46% of invocations compiled, 4.5% of work displaced.

## The two gaps

**1. `new` expressions are not compiled at all.** `M` opens with
`var t = new Float64Array(31)`. Allocating a scratch field element is the
first statement of essentially every tweetnacl field operation, so this one
construct is enough to keep the whole hot path on the tree-walker.

**2. Compound assignment to a member target is not compiled.**
`compiler.rs:295` accepts `Expression::Member` only under
`*op == AssignOp::Assign`; every other operator falls through to
`Err(CompileError::Unsupported("assign target"))` at `compiler.rs:317`. So
`o[i] = x` compiles but `o[i] += x`, `o[i] -= x` and `o[i] ^= t` do not —
which is the entire body of `car25519` and `sel25519`, and the accumulation
loop of `M`:

```js
function car25519(o) {                       // every statement is a compound
  for (var i = 0; i < 16; i++) {             // member assignment
    o[i] += 65536;
    var c = Math.floor(o[i] / 65536);
    o[(i+1)*(i<15?1:0)] += c - 1 + 37*(c-1)*(i===15?1:0);
    o[i] -= c * 65536;
  }
}
```

[...]

PR #605, opened 2026-09-05
docs(libtests): record the measured bytecode gate for tweetnacl sampling

pmatos:
## Summary

- Answers the open question in #361 by measuring it: the bytecode VM does **not**
  move tweetnacl-js's curve25519/Ed25519 arithmetic (1.00-1.07x, inside this
  host's run-to-run spread), because the three functions carrying 96% of the
  tree-walker's work all bail out of the compiler.
- Replaces the unevaluable deferral note ("revisit once the engine has a faster
  numeric path") in `scripts/libs/tweetnacl-js.sh` and `scripts/README.md` with
  the measured gate: the two concrete compiler gaps (#603) plus the ~17x
  multiplier the full vector counts actually need.
- Comment-only. The executable content of the config is byte-identical to HEAD.

## Measurements

jsse tree-walker vs `--bytecode`, min of 3 reps, ms/op, `nacl.min.js` 1.0.3
(the build the harness bundles), deterministic inputs:

| operation | tree-walker | `--bytecode` | speedup | Node 25 |
|---|---|---|---|---|
| `scalarMult.base` | 2982.20 | 2973.40 | 1.003x | 14.00 |
| `scalarMult` | 3010.20 | 2977.80 | 1.011x | 10.75 |
| `sign.detached` | 5184.67 | 5190.67 | 0.999x | 37.15 |
| `sign.detached.verify` | 10834.00 | 10112.67 | 1.071x | 69.75 |

Individual reps varied ±15% on this shared host and some `--bytecode` reps came
in *slower* than the default, so these are null results, not small wins.

## Why — from the `perf-counters` build (deterministic, load-immune)

In a `scalarMult.base` run under `--bytecode` the VM executes 5,584,177 ops
while 136,962,403 AST work units remain: it displaces **3.3%** of the work
despite compiling 29% of function *invocations*. These three functions record
**identical work-unit counts with and without `--bytecode`** — the direct
evidence the VM never reaches them. They are 96% of the tree-walker's total
work, and the 99% below is their share of what is *left* under `--bytecode`:

| minified | tweetnacl | share of AST work | bail |
|---|---|---|---|
| `X` | `M` (GF(2^255-19) multiply) | 74.07% | `expression:New` |
| `C` | `car25519` | 22.88% | `assign target` |
| `F` | `sel25519` | 2.06% | `assign target` |

`M` opens with `new Float64Array(31)` and `new` is not compiled at all;
`compiler.rs:295` accepts a member assignment target only under
`AssignOp::Assign`, so `o[i] += x` / `o[i] ^= t` — the whole body of `car25519`
and `sel25519` — fall through to `Unsupported("assign target")` at
`compiler.rs:317`. Filed as #603.

## Why this leaves issue 361 open

Per the triage decision on that issue, the bytecode path not helping means the
sampled harness stays in place as a long-term, perf-gated deferral, so #361
remains the tracker for exhaustive coverage. This PR only makes the gate
precise enough to act on. Note that #603 is necessary but not sufficient: the
VM's speedup on typed-array field arithmetic is currently **unmeasured** (no
phase here ran hot numeric code through the VM), and full counts need ~17x
(~6h projected vs ~22min sampled today, both projections from the same model).

~~The extrapolation model validates against the one figure already recorded in
the repo: it predicts 9.9 min for the fixed 200-iteration `scalarMult.base` KAT,
against the "~11min" in the existing config header.~~

**Retracted** — that was circular. The config's `~11min` entered in the same
commit as the `≈3.4s` per-op figure this work supersedes (200 × 3.4 s =
11.3 min), so it is that number restated, not an independent observation.
Agreeing with it only means agreeing with the measurement it came from. The
projection is **unvalidated**; the config comment is corrected to ~10 min and
the report's Method section now says so explicitly.

## Test plan

- [x] `./scripts/lint.sh` — clean (both clippy profiles)
- [x] `bash -n` + sourcing the config — parses; `LIB_REF`/`LIB_EXPECT_COUNT`/`LIB_TIMEOUT`/`LIB_ENTRY` unchanged
- [x] `shellcheck -S warning` findings byte-identical before/after — none introduced
- [x] No executable value changed: with trailing comments normalised away, the config is identical to `main` (one trailing inline comment was corrected, so this is no longer literally comment-only)
- [x] No Rust touched, so `cargo test` / test262 cannot regress — not re-run for a comment-only change

---

*Updated after review: the projection ratio was stated as ~18x but is 16.85x; the 99% work share was quoted without naming its population (it is the residual under `--bytecode`, not the tree-walker total of 95.7%). Corrected in `ded4a8e`.*


pmatos:
## Code review — `origin/main...HEAD` (xhigh, two-axis)

Fixed point `a0ef691`. Findings applied in `ded4a8e`.

### Standards

Documented sources checked: root `CLAUDE.md`, `scripts/README.md` house style,
sibling `scripts/libs/*.sh` headers, `docs/agents/*`. No `CONTRIBUTING.md`.

- **HARD — wrong arithmetic.** `scripts/libs/tweetnacl-js.sh` and
  `scripts/README.md` both stated the full-vs-sampled projection as "roughly
  18x". Recomputing from the same per-op costs: 363.08 min / 21.55 min =
  **16.85x**. Fixed to ~17x, and in #603 and the PR body which repeated it.
- **HARD — percentage quoted without its population.** "three functions hold 99%
  of the interpreter work" is their share of AST work *remaining under
  `--bytecode`*; of the tree-walker total it is **95.70%**. This is the same
  class of error the repo has been burned by before (`docs/perf/` counter
  accounting), where a figure is correct but its denominator is unstated.
  Restated as the identical work-unit counts with and without `--bytecode`,
  which is the direct observation rather than a derived ratio.
- **HARD — verb did not match the quantity.** "the VM executes ~3% of the work":
  3.34% is the work *displaced*; ops actually *executed* are 3.92%. Now
  "displaces", matching the issue comment.
- **JUDGEMENT — unsourced average.** "~1.03x on offer" was not derived from
  anything stated; replaced with the measured **1.00-1.07x** range, which the
  preceding sentence already lists.
- Verified and left alone: the `new` / `AssignOp::Assign` claims hold against
  `src/interpreter/bytecode/compiler.rs` at HEAD (member targets are accepted
  only under `AssignOp::Assign`, everything else reaches the `"assign target"`
  catch-all); `M`/`car25519`/`sel25519` attributions match the minified source;
  commit authorship complies with `CLAUDE.md`; no new shellcheck findings; the
  config's executable content stays byte-identical to HEAD.
- No baseline smells: the diff adds no code.

### Spec

Spec is #361's Ask plus the 2026-09-04 triage decision.

- The decision is a conditional — *"if it meaningfully closes the perf gap,
  raise/remove the sample caps ... If the bytecode path doesn't help enough ...
  leave the sampled harness in place as a long-term, perf-gated deferral"*. The
  measurement resolves it to the second branch, and the change takes exactly that
  branch: caps untouched, sampling note kept. **No MISSING findings.**
- **Not scope creep, on balance:** filing #603 records a finding rather than
  *"pursuing a dedicated fast path now"*, which is what the decision rules out;
  and #603 is general compiler coverage, not a tweetnacl-specific path. Editing
  `scripts/README.md` is beyond the letter of the "doesn't help" branch (docs are
  only mentioned in the "helps" branch) but is what makes a *"perf-gated"*
  deferral gateable.
- **Correct that this PR does not close #361** — the deferral branch keeps it as
  the tracker.
- The "unmeasured" caveat on what the VM would deliver post-#603 is applied
  consistently in all four venues (both files, the issue comment, #603).

**Summary — Standards: 4 findings (worst: the 18x → 17x arithmetic error, since a
wrong number in a long-lived comment is what a future reader would act on).
Spec: 0 findings.** All Standards findings fixed in `ded4a8e`; the three that had
propagated to #603, the PR body and the #361 comment are corrected there too.

pmatos:
## Second review round — sub-agent findings on `fef3f94`

The two axis agents reported after the first round of fixes had already landed.
Three of their findings were the ones already fixed in `ded4a8e` (the 18x
arithmetic, the unpinned 99% denominator, the unsourced ~1.03x) — independent
confirmation. **Four further findings were new and are now fixed in `ad7def9`.**

### Fixed

- **HARD (both axes) — the change contradicted prose ~26 lines above it.**
  `README.md:352-356` and `tweetnacl-js.sh:32-35` still asserted "100-3500x",
  "≈3.4s vs ≈35ms", "≈12s vs ≈76ms" and "~7h" against the newly added 140-400x,
  2.98s/14ms and ~6h. Both files stated two unqualified values for one quantity.
  The older paragraph is now rewritten around the current numbers and says
  explicitly that they supersede the ones taken when the harness landed, and
  that ~6h is the same order as the ~7h first estimated.
- **JUDGEMENT (standards) — header size outlier / duplicated prose.** The config
  header had grown 46 → 65 lines, over twice the next largest sibling
  (`css-tree.sh`, 31; median ~11), duplicating `README.md` near-verbatim so every
  fix needed making twice. The detailed narrative now lives in `README.md` only;
  the config keeps a compact statement of the gate, at 54 lines.
- **JUDGEMENT (spec) — the stricter bar quoted, the binding one omitted.** Both
  files gave only the 17x figure while `LIB_TIMEOUT="3600"` sits a few lines
  below and needs just ~6x. Both bars are now stated.
- **JUDGEMENT (spec) — over-reach in #603.** Its body claimed "the payoff is not
  limited to tweetnacl", presupposing the payoff its own Scope note calls
  unmeasured. Reworded to claim the *gap* generalises, not the payoff. Corrected
  on the issue.
- Also carried across: the note that several `--bytecode` reps came in slower
  than the default, which had been in the issue comment but not the files.

### Fixed outside the diff

- **MISSING (spec) — the deferral existed in prose but not in tracker state.**
  #361 still carried `agent-ready` ("Fully specified, ready for an AFK agent"),
  so it stayed dispatchable and the same measurement would get re-run. Label
  removed, with the reasoning on the issue. The `sym:*` labels were left alone —
  those are Symphonika's claim state, not triage state.

### Noted, not actioned

Neither axis found a spec defect in the substance: the triage conditional is
resolved on the correct branch, not closing #361 is faithful to the deferral,
and #603 is in scope rather than the "dedicated fast path" the decision ruled
out. Both independently verified the `compiler.rs:295` / `:317` / `:739`
mechanics and the `M`/`car25519`/`sel25519` attribution against HEAD.

**Round-2 summary — Standards: 1 HARD + 3 judgement, all fixed. Spec: 1 missing
(fixed on the issue) + the numeric findings already covered.** Worst issue was
the ~7h/~6h contradiction: exactly the stale-neighbour failure this repo has
been bitten by before, and I introduced it by adding correct numbers next to
superseded ones without reconciling them.

pmatos:
## `/simplify` pass — `4d449c3`

Quality-only pass over the diff (no correctness hunting). The load-bearing
finding was an **altitude** one, and it subsumed most of the others.

### Applied

**Altitude — this was documented at the wrong layer.** The measurement was
living as prose in a config header and a README section, encoding a perf gate a
human has to read and honour. `docs/perf/` is where this repo already keeps
exactly this artifact: a dated directory holding a write-up plus the counter
dumps behind it — `2026-08-26/mandreel-bytecode-work-share.md` (which `CLAUDE.md`
cites) and `2026-09-02/typescript-octane-bail-report.md` are the same genre, the
latter literally a bail report. Added `docs/perf/2026-09-05/` with the write-up,
both perf-counter dumps, the per-rep timings and the benchmark body.

**Reuse — one number had four homes.** Config header, README, #361 and #603 each
restated the same figures, so the two earlier correction rounds each had to be
applied in several places (and the PR body in a fifth). Now one home, with
pointers. This also reuses the established `docs/perf/` convention instead of
inventing a location.

**Simplification — scar tissue from two correction rounds.** The prose had
accreted clarifying asides bolted on to fix earlier imprecision. The per-op
figures are now tables in the perf doc, where a table beats a paragraph, and both
prose sites state the gate and stop.

**Efficiency — the evidence was unreproducible.** Every number here was backed by
binaries and bench files under `/tmp`, which this host wipes hourly; the
binaries were already gone by the time of this pass. The counter dumps and
benchmark are now committed, and the regenerated dumps are byte-identical to the
originals, which re-verifies the figures.

Net effect: the config header goes 65 → 51 lines (46 on `main`, down from a peak
of over twice the next-largest sibling config), and the README section shrinks
while saying the same thing.

### Skipped

- **Generalising the sampler.** `lib_prepare` stride-samples three vector files,
  and the README notes this is "the first sampled corpus in this harness". A
  shared cap/sampling knob in `run-library-tests.sh` would arguably be the right
  layer — but that changes harness behaviour for every library and is well
  outside this diff. Worth its own issue if wanted.
- **The `LIB_TIMEOUT` inline "~11min" figure.** My model predicts 9.9 min for
  that same KAT loop. These are different measurements (a real harness run vs a
  projection), not a contradiction, and the perf doc reconciles them explicitly —
  it is the model's validation check. Left as-is.
- **A `CLAUDE.md` pointer to the new perf doc.** The README and config both link
  it; adding a third entry point is churn.

All checks green; the config's executable content remains byte-identical to
`main`.

PR #606, opened 2026-09-05
fix(parser): calibrate MAX_PARSE_DEPTH per build profile

### LINKED ISSUE #599: Parser MAX_PARSE_DEPTH guard doesn't fire before native stack overflow in debug builds
Found while fixing up PR #598 (fuzz infra / lib.rs extraction), which added a
unit test (`lib_tests::run_on_engine_stack_borrows_non_static_data` in
`src/lib.rs`) asserting that deeply-nested (`"[".repeat(5000)`) array-literal
source hits the parser's catchable `MAX_PARSE_DEPTH` (4,000) guard rather than
overflowing the native stack, even when run on the 128 MiB `ENGINE_STACK_SIZE`
thread.

That assertion holds in release builds but **not** in debug builds: binary
search on a debug build of `jsse` shows the native stack overflows (SIGABRT,
"thread 'jsse-engine' has overflowed its stack") somewhere between 1,000 and
1,500 levels of `[` nesting — well below `MAX_PARSE_DEPTH`'s 4,000 threshold.
Debug-mode recursive-descent frames are apparently large enough (no inlining,
extra bookkeeping) that ~1,000+ levels alone exceed 128 MiB, before the depth
counter (`Parser::enter_recursion`, `src/parser/mod.rs`) ever gets a chance to
reject the input.

Repro (debug build):
```
for n in 1000 1500; do
  python3 -c "print('['*$n)" > /tmp/deep.js
  ./target/debug/jsse /tmp/deep.js   # n=1000 parses/errors cleanly (EOF), n=1500 SIGABRTs
done
```

**Why deferred:** out of scope for #598 (a fuzzing-infrastructure PR) — this is
a pre-existing engine characteristic (`MAX_PARSE_DEPTH`'s comment in
`src/parser/mod.rs` already documents it as calibrated against the "128 MiB
execution stack", i.e. release-mode headroom) that the new test happened to
surface for the first time, not a regression #598 introduced. Fixing it
properly means either giving the parser meaningfully smaller per-frame stack
usage in debug builds or making the guard threshold profile-aware, both
bigger changes than an autofix pass should make.

**Immediate mitigation (landed in #598):** the new test is
`#[cfg_attr(debug_assertions, ignore = "...")]`-gated so `cargo test` (debug
profile, the CI gate) doesn't SIGABRT; it still runs under a release test
build.

Related: `CLAUDE.md`'s Acorn section already documents the same overflow
shape for 8.17.0+'s parser stack-guard test (`"[".repeat(2000)`), tracked
there as "a jsse deep-recursion robustness gap tracked separately" — this
issue is that separate tracking.

pmatos:
## Summary

`MAX_PARSE_DEPTH` bounds recursive-descent depth in *units*, but the resource it
actually protects is native stack *bytes* — and an unoptimized frame costs ~10x
an optimized one. The single 4,000 limit was calibrated for release frames, so
on a debug build deeply nested source exhausted the 128 MiB engine stack
(SIGABRT) before the guard could raise its catchable error.

- Split the constant per profile: **4,000 release (unchanged), 400 debug**.
- Un-ignores the two deep-nesting tests in `src/lib.rs` that #598 had to gate off
  in debug, plus a new profile-agnostic parser regression test and the
  worst-case shapes in `tests/recursion-limit-parser.js`.

Fixes #599

## Calibration

Measured on the 128 MiB engine stack by disabling the guard and binary searching
the abort point over the shapes that spend the most native stack per depth unit
(`[`, `({a:`, `({a:[`, `` `${ ``, `[{a:`):

| profile | native capacity | limit | margin |
|---------|-----------------|-------|--------|
| release | ~37,000 units   | 4,000 | 9.3x   |
| debug   | ~3,800 units    | 400   | 9.4x   |

Worst shape was `[{a:` at 3,768 units in debug — i.e. debug capacity sat *below*
the old 4,000 limit, so the guard could never fire there. Each profile now sits
~9x below its own capacity.

For scale: test262's deepest nesting across all 53,690 files is 64 brackets
(~192 units), so neither limit is near real code. Note one source nesting level
costs more than one unit (each `[` spends 3).

## Notes

- **Release behaviour is unchanged.** Re-measured the guard threshold for 13
  nesting shapes against the pre-change binary — identical in every case, as
  expected since the `not(debug_assertions)` arm is still 4,000.
- The issue describes debug `cargo test` as the CI gate; CI actually runs
  `cargo nextest run --release` (`ci.yml:56`), so the `ignore` attributes were
  only suppressing *local* debug runs. Those now pass.
- `cargo-fuzz` builds are optimized *with* `debug_assertions`, so
  `parse_roundtrip` picks up the 400 limit — the conservative direction, and a
  parse error is already an expected outcome for that target. `differential`
  shells out to `target/release/jsse`, so the node cross-check is unaffected.
- **Out of scope, but found while verifying:** the interpreter's guards
  (`CALL_DEPTH_*`, `EVAL_DEPTH_LIMIT`) have the same miscalibration —
  `tests/recursion-limit-interpreter.js` and a deep left-nested expression both
  SIGABRT on a debug build today while passing in release. Filed separately.

## Test plan

- [x] `cargo test` (debug) — 647 passed, 0 failed; the two previously-ignored tests now run
- [x] `cargo test --release` — 647 passed, 0 failed (`cargo nextest` not installed locally)
- [x] `./scripts/lint.sh` — clean (fmt + clippy + clippy --features perf-counters)
- [x] `run-custom-tests.py` — 11/11
- [x] test262 — 99,568/99,568 (100%), **0 regressions**
- [x] test262-extra — 308/308 default and `--bytecode`, 0 regressions
- [x] 21 nesting shapes verified to raise the catchable error (never SIGABRT) on the debug binary

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -390,7 +390,7 @@ pub(crate) struct Interpreter {
 
 /// Soft JS call-depth limit: crossing it throws a catchable `RangeError:
 /// Maximum call stack size exceeded`. Sits well below the native capacity of
-/// the 128 MiB execution stack the engine runs on (see `main.rs`), yet far
+/// the 128 MiB execution stack the engine runs on (see `lib.rs`), yet far
 /// above the depth any real program reaches (the old 8 MiB main stack held
 /// every passing test, i.e. depths under ~1500).
 pub(crate) const CALL_DEPTH_SOFT_LIMIT: usize = 4_000;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,7 +86,7 @@ mod lib_tests {
         // `ast::Program` holds `Rc`s and isn't `Send`, so (matching how a real
         // fuzz target must behave) the parse result is consumed entirely
         // inside the closure; only the `bool` verdict crosses the thread.
-        let owned = "[".repeat(5000);
+        let owned = "[".repeat(parser::MAX_PARSE_DEPTH as usize * 2);
         let borrowed: &str = &owned;
         let is_err = run_on_engine_stack(move || {
             parser::Parser::new(borrowed)
@@ -116,6 +116,6 @@ mod lib_tests {
 
     #[test]
     fn fuzz_parse_bytes_deep_nesting() {
-        fuzz_parse_bytes("[".repeat(5000).as_bytes());
+        fuzz_parse_bytes("[".repeat(parser::MAX_PARSE_DEPTH as usize * 2).as_bytes());
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,13 +82,6 @@ mod lib_tests {
     }
 
     #[test]
-    #[cfg_attr(
-        debug_assertions,
-        ignore = "MAX_PARSE_DEPTH's headroom below ENGINE_STACK_SIZE is calibrated \
-                  for release-mode stack frames; debug-mode recursive-descent frames \
-                  are large enough that this depth overflows the native stack well \
-                  before the depth guard fires. See jsse#599."
-    )]
     fn run_on_engine_stack_borrows_non_static_data() {
         // `ast::Program` holds `Rc`s and isn't `Send`, so (matching how a real
         // fuzz target must behave) the parse result is consumed entirely
@@ -122,13 +115,6 @@ mod lib_tests {
     }
 
     #[test]
-    #[cfg_attr(
-        debug_assertions,
-        ignore = "MAX_PARSE_DEPTH's headroom below ENGINE_STACK_SIZE is calibrated \
-                  for release-mode stack frames; debug-mode recursive-descent frames \
-                  are large enough that this depth overflows the native stack well \
-                  before the depth guard fires. See jsse#599."
-    )]
     fn fuzz_parse_bytes_deep_nesting() {
         fuzz_parse_bytes("[".repeat(5000).as_bytes());
     }
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -91,6 +91,10 @@ pub(crate) struct Parser<'a> {
 /// raising this error in debug builds — jsse#599.
 #[cfg(not(debug_assertions))]
 const MAX_PARSE_DEPTH: u32 = 4_000;
+/// Debug counterpart of [`MAX_PARSE_DEPTH`], ~10x lower for ~10x fatter frames.
+/// Reaching either limit costs about the same ~14 MB of stack, so both assume
+/// the engine stack — a parser driven on a smaller thread can still overflow
+/// first, in either profile.
 #[cfg(debug_assertions)]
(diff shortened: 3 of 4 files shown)


pmatos:
Follow-up for the interpreter-side siblings noted above: #607.

pmatos:
## Two-axis code review (`origin/main...HEAD`, xhigh) — findings applied in c9ed12f

**Standards axis: 0 documented-standard violations**, 5 judgement-call smells.
**Spec axis: 0 missing requirements**, 1 real defect, 1 minor scope note.

### Fixed

1. **False claim in a doc comment** (both axes flagged this; Spec called it the one real defect). The regression test's doc said its five shapes are "the most stack-hungry ones the parser has, so they bound every other shape". Not true: source that consumes native stack *without advancing `parse_depth`* is outside the guard entirely. Verified —

   | source | debug | release |
   |---|---|---|
   | `function f(){return a` + `.b`x200,000 + `}` | rc=134 (SIGABRT) | rc=0 |
   | `var x = ` + `1+`x200,000 + `1;` | rc=134 (SIGABRT) | rc=1 (catchable) |

   Pre-existing and unrelated to this change — `parse_depth` never advanced for these shapes in *either* profile, so the guard never applied; release survives only because its frames are ~10x smaller. The comment now says the shapes bound every shape *the guard covers*, and names the excluded families. Recorded on #607.

2. **A second inaccurate claim:** `nesting_below_the_limit_still_parses` was documented as testing "an order of magnitude past test262's deepest (64 brackets)" while testing exactly 64 — and 640 would *not* parse in debug (400 units caps array nesting at ~133 levels). Corrected rather than "fixed" by changing the test, since 64 is the right value.

3. **Undocumented debug arm** — the 17-line doc attached only to the release constant. The debug arm now carries its own note, including that reaching the guard costs ~13.6 MB (debug) / ~13.8 MB (release), so both limits assume the engine stack.

4. **Duplicated Code** — the `run_on_engine_stack` + non-`Send` `Program` dance appeared in both new tests; folded into one `parse_on_engine_stack` helper. Added the `[...` spread shape so the Rust list is a superset of the JS one (they were diverging).

5. **Stale cross-reference** — `src/interpreter/mod.rs:393` still pointed at `main.rs` for `ENGINE_STACK_SIZE`, which lives in `lib.rs`. (The report cited two sites; only one exists.)

### Not actioned

- **Scope note (Spec):** the four extra `mustThrow` cases in `tests/recursion-limit-parser.js` weren't asked for by #599. Keeping them — they are the shapes that actually reach the native limit first, which is the regression worth guarding.

### Re-validation after the fixes

`cargo test` debug 647/0 · `cargo test --release` 647/0 · `./scripts/lint.sh` clean · custom tests 11/11 · 8 nesting shapes still GUARD (never SIGABRT) in debug · release trip points unchanged (`[` 1333, `({a:` 667, `({a:[` 445, `` `${ `` 1333).

PR #622, opened 2026-09-10
refactor(iterators): extract with_array_elements/_mut, retire 9 borrow preambles

### LINKED ISSUE #610: Extract with_array_elements/with_array_elements_mut and retire the 9 hand-written borrow preambles in iterators.rs
## Motivation

`src/interpreter/builtins/iterators.rs` spells out the same "JsValue -> array elements" borrow sequence nine times:

```rust
let buffer_id = buffer
    .as_object_id()
    .expect("chunks buffer must be an Array object");
let cell = interp.get_object_cell_expect(buffer_id);
let mut obj = cell.borrow_mut();
let elements = obj
    .array_elements_mut()
    .expect("chunks buffer must have Array elements");
```

Sites: `chunks` (~2685, 2698, 2712, 2726), `windows` (~2843, 2868), and `RootedPair::slots`/`get`/`set` added by #604. There is no shared combinator today — `get_object_cell_expect` (`interpreter/mod.rs:1973`) and `array_elements`/`array_elements_mut` (`interpreter/types.rs:2175`/`2179`) are the raw pieces, and a repo-wide grep finds no wrapper.

## Cost

Nine hand-written borrow sequences restating the same `expect` invariant ("this JsValue is an engine-allocated Array"). Anyone changing `ArrayData`'s storage or the write-barrier discipline has to touch all nine.

## Proposed shape

Inherent methods on `Interpreter`, next to `get_object_cell_expect`:

```rust
pub(crate) fn with_array_elements<R>(&self, array: &JsValue, f: impl FnOnce(&Vec<JsValue>) -> R) -> R;
pub(crate) fn with_array_elements_mut<R>(&self, array: &JsValue, f: impl FnOnce(&mut Vec<JsValue>) -> R) -> R;
```

Both borrow-check (the `RefMut` lives only inside the call), `_mut` keeps the generational write barrier via `borrow_mut()`, and no call site re-borrows `interp` inside `f`. `RootedPair::get`/`set`/`clear` become one-liners and `slots()` disappears.

## Why not in #604

Six of the nine sites are in `chunks`/`windows`, outside that PR's diff; adding the API for only the three new sites would leave the pattern half-converted, which is worse than either end state. Raised by a reuse review of #604 and deferred deliberately.

Note this is orthogonal to #609 (promoting the rooted-slot container itself to `gc.rs`) but the two touch the same code — worth sequencing.

pmatos:
## Summary

- Add `Interpreter::with_array_elements` / `with_array_elements_mut` next to `get_object_cell_expect`, consolidating the `as_object_id -> get_object_cell_expect -> borrow(_mut) -> array_elements(_mut)` chain into a single call.
- Convert all 9 hand-written borrow preambles in `iterators.rs` to use the new helpers: 4 in `chunks()`, 2 in `windows()`, and `RootedPair::get`/`set` (its `slots()` helper is now unneeded and removed).
- `_mut` still goes through `ObjectHandle::borrow_mut()`, so the generational write barrier is unaffected; the `RefMut`/`Ref` never escapes the closure.

Fixes #610

## Test plan

- [x] `cargo build --release` — clean
- [x] `./scripts/lint.sh` — clean (rustfmt, clippy, clippy perf-counters)
- [x] `uv run python scripts/run-custom-tests.py` — 11/11 passing
- [x] `uv run python scripts/run-test262.py test262/test/built-ins/Iterator/ test262/test/staging/sm/Iterator` — 1656/1656 passing, 0 regressions
- [x] `uv run python scripts/run-test262.py` (full suite) — 99907/99907 passing, 0 regressions

### DIFF
--- a/src/interpreter/builtins/iterators.rs
+++ b/src/interpreter/builtins/iterators.rs
@@ -2675,12 +2675,9 @@ impl Interpreter {
                             Ok(Some(value)) => {
                                 let values = interp.with_array_elements_mut(&buffer, |elements| {
                                     elements.push(value);
-                                    (elements.len() == chunk_size).then(|| elements.clone())
+                                    (elements.len() == chunk_size).then(|| std::mem::take(elements))
                                 });
                                 if let Some(values) = values {
-                                    interp.with_array_elements_mut(&buffer, |elements| {
-                                        elements.clear()
-                                    });
                                     let chunk = interp.create_array(values);
                                     return Completion::Normal(
                                         interp.create_iter_result_object(chunk, false),
@@ -2689,15 +2686,13 @@ impl Interpreter {
                             }
                             Ok(None) => {
                                 state_next.borrow_mut().3 = false;
-                                let values = interp
-                                    .with_array_elements(&buffer, |elements| elements.clone());
+                                let values =
+                                    interp.with_array_elements_mut(&buffer, std::mem::take);
                                 if values.is_empty() {
                                     return Completion::Normal(
                                         interp.create_iter_result_object(JsValue::UNDEFINED, true),
                                     );
                                 }
-                                interp
-                                    .with_array_elements_mut(&buffer, |elements| elements.clear());
                                 let chunk = interp.create_array(values);
                                 return Completion::Normal(
                                     interp.create_iter_result_object(chunk, false),
--- a/src/interpreter/builtins/iterators.rs
+++ b/src/interpreter/builtins/iterators.rs
@@ -369,31 +369,17 @@ impl RootedPair {
         self.0.clone()
     }
 
-    fn slots<'a>(&self, interp: &'a Interpreter) -> &'a ObjectHandle {
-        interp.get_object_cell_expect(
-            self.0
-                .as_object_id()
-                .expect("rooted pair must be an Array object"),
-        )
-    }
-
     fn get(&self, interp: &Interpreter) -> Option<(JsValue, JsValue)> {
-        let cell = self.slots(interp);
-        let obj = cell.borrow();
-        let elements = obj
-            .array_elements()
-            .expect("rooted pair must have Array elements");
-        (!elements[0].is_undefined()).then(|| (elements[0].clone(), elements[1].clone()))
+        interp.with_array_elements(&self.0, |elements| {
+            (!elements[0].is_undefined()).then(|| (elements[0].clone(), elements[1].clone()))
+        })
     }
 
     fn set(&self, interp: &Interpreter, iterator: JsValue, next_method: JsValue) {
-        let cell = self.slots(interp);
-        let mut obj = cell.borrow_mut();
-        let elements = obj
-            .array_elements_mut()
-            .expect("rooted pair must have Array elements");
-        elements[0] = iterator;
-        elements[1] = next_method;
+        interp.with_array_elements_mut(&self.0, |elements| {
+            elements[0] = iterator;
+            elements[1] = next_method;
+        });
     }
 
     fn clear(&self, interp: &Interpreter) {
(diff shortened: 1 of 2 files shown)


src/interpreter/mod.rs
```diff
@@ -1974,6 +1974,42 @@ impl Interpreter {
...
+        array: &JsValue,
+        f: impl FnOnce(&Vec<JsValue>) -> R,
+    ) -> R {
+        let id = array
+            .as_object_id()
+            .expect("with_array_elements: value must be an Array object");
+        let obj = self.get_object_cell_expect(id).borrow();
+        let elements = obj
+            .array_elements()
+            .expect("with_array_elements: object must have Array elements");
```
pmatos, line 1990: Cleanup, not a bug: the six former call sites each had a distinct `.expect(...)` message (`"rooted pair must be an Array object"`, `"chunks buffer must have Array elements"`, `"windows buffer must have Array elements"`, etc.). Collapsing them into these two generic messages means a future invariant violation (e.g. a caller accidentally roots a non-Array value) will panic with `"with_array_elements: object must have Array elements"` instead of naming which buffer/slot was wrong — a minor debuggability regression if that panic path is ever hit. Not blocking; a small win would be an optional `&str` label parameter if this is expected to grow more callers.

src/interpreter/builtins/iterators.rs
```diff
@@ -2687,25 +2673,14 @@ impl Interpreter {
...
-                                let buffer_id = buffer
-                                    .as_object_id()
-                                    .expect("chunks buffer must be an Array object");
-                                let values = {
-                                    let cell = interp.get_object_cell_expect(buffer_id);
-                                    let mut obj = cell.borrow_mut();
-                                    let elements = obj
-                                        .array_elements_mut()
-                                        .expect("chunks buffer must have Array elements");
+                                let values = interp.with_array_elements_mut(&buffer, |elements| {
```
pmatos, line 2676: `chunks()`'s `Ok(Some(value))` and `Ok(None)` branches each `elements.clone()` the full buffer and then make a **second**, separate `with_array_elements_mut` call just to `.clear()` it — two borrows and an extra O(n) allocation where one would do. Since the helper already hands you `&mut Vec<JsValue>`, `std::mem::take(elements)` returns the owned Vec and leaves an empty one behind in a single borrow:

```rust
Ok(Some(value)) => {
    let values = interp.with_array_elements_mut(&buffer, |elements| {
        elements.push(value);
        (elements.len() == chunk_size).then(|| std::mem::take(elements))
    });
    if let Some(values) = values {
        let chunk = interp.create_array(values);
        return Completion::Normal(interp.create_iter_result_object(chunk, false));
    }
}
Ok(None) => {
    state_next.borrow_mut().3 = false;
    let values = interp.with_array_elements_mut(&buffer, std::mem::take);
    if values.is_empty() {
        return Completion::Normal(interp.create_iter_result_object(JsValue::UNDEFINED, true));
    }
    let chunk = interp.create_array(values);
    return Completion::Normal(interp.create_iter_result_object(chunk, false));
}
```

(`windows()` doesn't need this — it never clears, only slides via `remove(0)`.) I verified this variant locally: builds clean, `./scripts/lint.sh` passes, and all 156 `chunks`/`windows` test262 scenarios plus the full `cargo test --release --lib` suite (669 passed) still pass.

src/interpreter/mod.rs
```diff
@@ -1974,6 +1974,42 @@ impl Interpreter {
         self.objects.get_cell_expect(id)
     }
 
+    /// Borrow `array`'s element vector immutably for the duration of `f`.
+    /// Panics if `array` is not an engine-allocated Array object.
+    pub(crate) fn with_array_elements<R>(
```
pmatos, line 1979: Non-blocking architecture note: this correctly retires the 9 preambles in `iterators.rs`, but the same hand-rolled `as_object_id()` → `get_object_cell_expect()` → `borrow()/borrow_mut()` → `array_elements()/array_elements_mut()` chain still appears at 50+ other sites (`property.rs`, `eval.rs`, `builtins/array.rs`, `gc.rs`, `helpers.rs`, `builtins/mod.rs`, `builtins/intl/segmenter.rs`, and the analogous map/set-data pattern in `builtins/collections.rs`). Two idioms for the same operation now coexist, so a future contributor touching those files may not know `with_array_elements[_mut]` exists. Not asking to fix all of them here — just flagging that the win is currently scoped to one file, in case a follow-up issue to migrate the rest (or generalize the helper past Array specifically) is worth filing.

pmatos (commented):
xhigh-effort review (10 parallel finder angles + 1-vote verification + gap sweep). This is a clean, behavior-preserving refactor: borrow scopes are equivalent to the pre-refactor inline code at every one of the 9 converted call sites, the write barrier (`remember_if_old` via `ObjectHandle::borrow_mut`) is preserved, no double-borrow/reentrancy risk, `RootedPair::slots()` has zero remaining callers after removal, and no correctness bug survived any of the 10 finder angles. Verified independently: `cargo build --release` clean, `./scripts/lint.sh` clean (rustfmt + clippy, both profiles), `cargo test --release --lib` 669 passed, full `chunks`/`windows` test262 suite (156 scenarios) 100% pass, custom test suite 11/11 pass.

Three non-blocking cleanup notes left inline: (1) generic panic messages lose per-call-site specificity, (2) `chunks()`'s clone-then-separately-clear pattern can collapse to one `std::mem::take`-based borrow (verified fix included, all tests still green), (3) an architecture note that ~50 other call sites elsewhere in the interpreter still hand-roll the same borrow preamble this PR retires only in `iterators.rs`. None of these block merging.

Issue #465, opened 2026-08-19
Persistent async resolver roots share the truncatable gc_temp_roots stack and are dropped by enclosing frames


pmatos:
## Summary

`gc_temp_roots` carries two kinds of root with incompatible lifetimes:

- **frame-scoped temporaries**, reclaimed in bulk by `gc_unroot_frame(frame)` — which is just `self.gc_temp_roots.truncate(frame)` (`src/interpreter/mod.rs:1003-1005`);
- **persistent async roots**, which must survive until an async completion settles — the promise resolver rooted by `$262.agent.getReportAsync` (`src/interpreter/mod.rs:699`, released at `mod.rs:748`) and by `Atomics.waitAsync` (`src/interpreter/builtins/atomics.rs:576-578`, released at `atomics.rs:642`).

Because both live on one stack, **any enclosing frame truncation discards the persistent root.** A subsequent collection can then free the resolver, and the promise stays pending forever.

## The codebase already knows this, and two places defend against it

`call_function_inner_impl` deliberately unroots operand-by-operand instead of truncating, and says why:

```rust
// src/interpreter/eval.rs:12712-12716
// Unroot only what we pushed. A bulk truncate would also
// drop persistent roots pushed by the builtin itself
// (e.g. Atomics.waitAsync / $262.agent.getReportAsync
// root resolve_fn for an async completion that runs
// after this call returns).
```

`array_set_length` does the same, for the same stated reason (`src/interpreter/property.rs:422-427`).

## But the outermost call frame defeats it

`eval_call` takes its marker *before* the call and bulk-truncates *after* it:

```rust
// src/interpreter/eval.rs:5433
let gc_frame = self.gc_root_frame();
...
// src/interpreter/eval.rs:5519-5520
let result = self.call_function(&func_val, &this_val, &evaluated_args);
self.gc_unroot_frame(gc_frame);
```

So the careful cleanup at `eval.rs:12717-12720` is undone one level up. A plain

```js
var p = $262.agent.getReportAsync();
```

pushes the resolver root during `call_function` at `eval.rs:5519` and loses it at `eval.rs:5520`, the instant the builtin returns — with no private fields, no accessors, and nothing exotic in the picture. The same applies to the ~40 other `gc_unroot_frame` sites in `eval.rs`.

## Why the root is load-bearing

[...]

PR #624, opened 2026-09-10
refactor(interpreter): migrate yield* delegation to with_gc_root_scope

pmatos:
## Summary

- Migrates the `yield*` delegation loop in `eval.rs` (`Expression::Yield` with `delegate == true`) to `with_gc_root_scope`, collapsing 10 hand-threaded `gc_unroot_frame` teardowns (one per abrupt exit across `IteratorNext` throwing, an awaited rejection, `done`/`value` getters throwing, and each of the `next`/`return`/`throw` resume kinds) into the combinator's single bulk-unroot. Replaces the raw `gc_temp_roots.push(o.id)` bypass with `gc_root_value(&iterator)`, retiring one of the two seam bypasses PR #595 flagged (`eval.rs:1066`/`:4324`).
- This is the concrete first slice of PR #595's proposed `gc-root-scope-guard-eval` follow-up, and #331 (filed before the generator state-machine transform existed) named it as the prototype target. **Reachability update, added after review**: as of this branch, ordinary `yield*` execution goes through `generator_runtime.rs`'s independently-implemented, already-correctly-GC-rooted `delegated_iterator` state machine instead of this loop — see "Notes for reviewers" below and jsse#625. The migration itself is still correct, tested, and harmless; treat it as hygiene on a legacy/fallback path rather than a fix to code most `yield*` calls execute today.
- Adds `docs/adr/2026-09-10-2014-gc-root-scope-guard.md`, formalizing PR #595's "Proposed ADR" (closure combinator, not an RAII `Drop`-guard) now that it has a second real call site, and recording why the remaining raw-rooting sites (`destructure_array_assignment`/`bind_pattern`'s identity-based unrooting, `Array.fromAsync`'s `from_async_gc_root`, and the `exec.rs`/`regexp.rs` mechanical-sweep stragglers) are deliberately not touched here.

Refs #331 — this is one slice of a larger staged backlog (RAII-vs-combinator was decided by #595; splitting frame-scoped from identity-scoped roots, and root-stack balance assertions, remain open). Not closing the issue.

## Notes for reviewers

- **A prior autonomous run on this issue cited issue #465 / commit `03df6fe`** as the documented reason `destructure_array_assignment`/`bind_pattern` are excluded from this migration. I verified that citation while writing the ADR: #465 (closed) is unrelated — it's about promise-resolver roots (`Atomics.waitAsync`/`$262.agent.getReportAsync`) being dropped by an enclosing frame's bulk truncation, not about destructuring. The underlying technical claim (those sites use per-value `gc_unroot_value` identity removal, not frame truncation, so migrating them risks unrooting a value a sibling branch still needs) checks out from reading the code directly, so the exclusion stands — the ADR just doesn't repeat the wrong citation.
- **A `/code-review xhigh --fix --comment` pass raised, then largely resolved through its own back-and-forth, a substantive question: is the loop this PR touches reachable at all?** Short answer, confirmed independently with direct instrumentation (a temporary `eprintln!` in the exact branch, not part of this diff): **no construct I could build reaches it.** Ordinary `yield*` — bare, inside `while`/`do-while`/nested-`while` loops, `try/finally`, `switch`, `for-in`, labeled `continue`, as a binary/call/ternary/array-literal operand, and in an async generator — all go through `generator_runtime.rs`'s `StateTerminator::Yield { is_delegate: true, .. }` arm, which suspends via `delegated_iterator` directly (spec §14.4.14) without calling back into `eval_expr`. There's a documented `InlineYield` fallback (`generator_runtime.rs:4301-4304`) that *would* reach this PR's code for control flow the transform can't decompose, but nothing tried triggers it. Filed as jsse#625 to track whether that fallback is ever live and whether `CLAUDE.md`'s "replay-based" generator description needs updating.
- **This also means the GC-forcing `test262-extra/` regression test I'd attempted and dropped earlier (getter-based `done` check calling `$262.gc()`) was never going to discriminate** — not just because `GeneratorContext::delegated_iterator`'s tracing is a broader safety net than I'd first assumed, but because the code under test isn't in the execution path for that repro at all. Consistent with dropping it rather than shipping a non-discriminating test.
- Given the above, the value of this PR is: (a) the `with_gc_root_scope` migration is still a correct, tested simplification of whatever *does* execute this loop, if anything ever does, and (b) the ADR now documents the seam decision accurately for the next person who touches this file, including the reachability caveat. If jsse#625 concludes the legacy `self.generator_context`/`exec_statements` yield* path is fully dead, this branch (and its sibling plain-`yield` code) becomes a deletion candidate rather than a migration target — noted there, not resolved here.

## Test plan

- [x] `cargo build --release`
- [x] `cargo test --release` — 670 + 6 passing, 0 failed
- [x] `./scripts/lint.sh` — rustfmt, clippy (default + perf-counters), all clean
- [x] `uv run python scripts/run-test262.py test262/test/language/expressions/yield/ test262/test/language/statements/generators/ test262/test/language/expressions/async-generator/ test262/test/language/statements/async-generator/ -j 32` — 2,435/2,435, 0 regressions
- [x] `uv run python scripts/run-test262.py -j 32` (full suite) — 99,907/99,907 (100%), 0 regressions
- [x] `uv run python scripts/run-test262.py test262-extra/ -j 32` — 316/316

### DIFF
--- a/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
+++ b/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
@@ -40,8 +40,8 @@ zero real adapters in `array.rs`).
   `gc_frame_next` held simultaneously), which correctly keep the raw
   primitive.
 - **`eval.rs`'s `yield*` delegation** (this change): the concrete first slice
-  of the `gc-root-scope-guard-eval` follow-up PR #595 proposed. This is also
-  the control-flow-heavy path #331 itself named as the prototype target — 11
+  of the `gc-root-scope-guard-eval` follow-up PR #595 proposed, and the
+  control-flow-heavy path #331 itself named as the prototype target — 10
   `gc_unroot_frame` call sites across the iterator-protocol loop, one per
   abrupt exit (`IteratorNext` throwing, an awaited rejection, `done`/`value`
   getters throwing, and each of `next`/`return`/`throw` resume kinds)
@@ -49,6 +49,28 @@ zero real adapters in `array.rs`).
   `if let Some(o) = iterator.as_object_id() { self.gc_temp_roots.push(o.id) }`
   bypass at this site is replaced with `gc_root_value(&iterator)`, retiring
   one of the two seam bypasses PR #595 flagged (`eval.rs:1066`/`:4324`).
+
+  **Reachability caveat, found during review of this ADR.** #331 described
+  this loop as *the* control-flow-heavy `yield*` path when it was filed.
+  Today, ordinary `yield*` execution goes through `generator_runtime.rs`'s
+  state-machine `StateTerminator::Yield { is_delegate: true, .. }` arm
+  instead — a separate, independently-suspending implementation (spec
+  §14.4.14) that stores `delegated_iterator` on the generator object rather
+  than calling back into `eval_expr`. This `eval.rs` loop is reached only
+  through the documented `InlineYield` fallback
+  (`generator_runtime.rs:4301-4304`: "any `Completion::Yield` from
+  `exec_statements` ... came from a loop body or complex control flow that
+  isn't decomposed by the state machine transformer"). Direct instrumentation
+  of this exact branch found zero hits across a bare `yield*`, `while`/
+  `do-while`/nested-`while` loops, `try`/`finally`, `switch`, `for-in`,
+  labeled `continue`, `yield*` as a binary/call/ternary/array-literal operand,
+  and an async generator — so no construction found so far exercises it. The
+  migration is still correct and harmless (identical behavior on every exit
+  path, full test262 green), but its practical value is unconfirmed rather
+  than the "control-flow-heavy path" framing alone would suggest; see
+  jsse#625 for the open question of whether/when this fallback fires and
+  whether the legacy `self.generator_context`-driven branch is still load-
+  bearing.
 
 ## Deliberately not migrated
 
--- a/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
+++ b/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
@@ -17,9 +17,13 @@ whole-body, single-frame native temp-root scoping. It is not a Drop-guard:
 `gc_temp_roots` stays a plain `Vec<u64>`, so there is no interior-mutability
 borrow tax (`RefCell`) on the GC hot path (`gc_root_value`, called from every
 allocation-adjacent site in the interpreter). The raw `gc_root_frame`/
-`gc_unroot_frame` primitive is retained for frames that genuinely nest or
-interleave across early exits — the one case the combinator structurally
-cannot express, since it owns exactly one depth marker.
+`gc_unroot_frame` primitive is retained for the one shape the combinator
+cannot express: two frames alive at once where an *inner* one must truncate
+independently while the *outer* one stays open across repeated early exits
+(`Array.from`'s nested `gc_frame`/`gc_frame_next`, see below). Plain LIFO
+nesting — one `with_gc_root_scope` call inside another, each owning its own
+local frame marker on the same stack — composes fine and is not what this
+primitive is reserved for.
 
 An RAII `Drop`-guard is added only if a genuinely interleaved case with two
 or more real adapters later surfaces, decided where the variation is
--- a/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
+++ b/docs/adr/2026-09-10-2014-gc-root-scope-guard.md
@@ -1,0 +1,95 @@
+# GC temp-root scoping: a closure combinator, not an RAII guard
+
+Issue #331 asked whether the GC temp-root frame idiom — `let f =
+self.gc_root_frame(); … self.gc_unroot_frame(f);`, paired and placed by hand
+on every exit path — should become an RAII `Drop`-guard, so a forgotten
+teardown (premature collection or a root leak) can no longer happen by
+omission. #290 first reconciled the one-line `gc_root_value(&x)` idiom; this
+records the decision for the frame-scoping half of that question, now that
+`with_gc_root_scope` has a second call site beyond its `array.rs` origin.
+
+## Decision
+
+`Interpreter::with_gc_root_scope(|interp| …)` — a closure combinator that
+captures the current temp-root depth, runs `body`, and bulk-unroots on every
+exit path the closure takes (tail, early `return`, `?`) — is the seam for
+whole-body, single-frame native temp-root scoping. It is not a Drop-guard:
+`gc_temp_roots` stays a plain `Vec<u64>`, so there is no interior-mutability
+borrow tax (`RefCell`) on the GC hot path (`gc_root_value`, called from every
+allocation-adjacent site in the interpreter). The raw `gc_root_frame`/
+`gc_unroot_frame` primitive is retained for frames that genuinely nest or
+interleave across early exits — the one case the combinator structurally
+cannot express, since it owns exactly one depth marker.
+
+An RAII `Drop`-guard is added only if a genuinely interleaved case with two
+or more real adapters later surfaces, decided where the variation is
+observed rather than pre-committed here. PR #595's design-it-twice review
+(`.architecture/reviews/2026-09-04-gc-root-scope-guard.md`) scored this
+against an RAII guard (Design B, rejected: forces `Rc<RefCell<Vec<u64>>>`,
+touches `gc.rs` and ~15 functions across six files for a capability
+`array.rs` never used) and a value-taking `with_gc_rooted(&[&JsValue], …)`
+variant (Design C, runner-up: its slice seals a ≥2-root variation that had
+zero real adapters in `array.rs`).
+
+## Applied so far
+
+- **`array.rs`** (PR #595): 8 whole-body natives — `concat`, `slice`, `map`,
+  `filter`, `splice`, `flat`, `flatMap`, `Array.from`'s array-like path —
+  collapsing 71 hand-threaded `gc_unroot_frame` teardowns to 9. The 9
+  remaining are `Array.from`'s nested iterator frames (`gc_frame` +
+  `gc_frame_next` held simultaneously), which correctly keep the raw
+  primitive.
+- **`eval.rs`'s `yield*` delegation** (this change): the concrete first slice
+  of the `gc-root-scope-guard-eval` follow-up PR #595 proposed. This is also
+  the control-flow-heavy path #331 itself named as the prototype target — 11
+  `gc_unroot_frame` call sites across the iterator-protocol loop, one per
+  abrupt exit (`IteratorNext` throwing, an awaited rejection, `done`/`value`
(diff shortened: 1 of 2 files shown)


src/interpreter/eval.rs
```diff
@@ -1058,83 +1058,58 @@ impl Interpreter {
...
-                            match self.await_value(&next_result) {
-                                Completion::Normal(v) => v,
-                                Completion::Throw(e) => {
-                                    self.gc_unroot_frame(gc_frame);
-                                    return Completion::Throw(e);
-                                }
-                                other => {
-                                    self.gc_unroot_frame(gc_frame);
-                                    return other;
+                    self.with_gc_root_scope(|this| {
```
pmatos, line 1061: **Finding: the refactored `yield*` delegation loop is unreachable dead code for every real generator.**

Every `function*`/`async function*` call always constructs `IteratorState::StateMachineGenerator`/`StateMachineAsyncGenerator` (see `eval.rs` ~5909 and ~5719, via `transform_generator`/`transform_async_generator`, both of which return a bare `GeneratorStateMachine` unconditionally — no bail path). `Generator.prototype.next/return/throw` (`builtins/iterators.rs` ~4113-4180) and `AsyncGenerator.prototype.next` (`generator_runtime.rs:6072`, `async_gen_enqueue`) explicitly detect the state-machine kind and route to `generator_next_state_machine`/`async_gen_enqueue` instead. `self.generator_context` — the field this loop reads at line 1088/1122 — is only ever set to `Some(...)` inside the legacy `generator_next`/`async_generator_next` (generator_runtime.rs:71, 6133), and those functions are only reached for an object whose kind is the plain (non-state-machine) `IteratorState::Generator`/`AsyncGenerator` — a kind nothing in the codebase ever constructs from `SuspendedStart`.

So this PR's GC-rooting fix, and the ADR's whole 'abrupt exit' risk analysis, apply to code no JS program can trigger. The cited '2,435/2,435 yield*/generator/async-generator test262 scenarios, 0 regressions' validates `generator_runtime.rs`'s separate, independently-implemented (and already correctly GC-rooted via `delegated_iterator`) yield*-delegation, not this diff. Worth confirming intentionally (e.g. is this legacy tree-walker generator path slated for deletion?) rather than continuing to invest ADR-level design rigor in it.

docs/adr/2026-09-10-2014-gc-root-scope-guard.md
```diff
@@ -0,0 +1,95 @@
...
+
+- **`array.rs`** (PR #595): 8 whole-body natives — `concat`, `slice`, `map`,
+  `filter`, `splice`, `flat`, `flatMap`, `Array.from`'s array-like path —
+  collapsing 71 hand-threaded `gc_unroot_frame` teardowns to 9. The 9
+  remaining are `Array.from`'s nested iterator frames (`gc_frame` +
+  `gc_frame_next` held simultaneously), which correctly keep the raw
+  primitive.
+- **`eval.rs`'s `yield*` delegation** (this change): the concrete first slice
+  of the `gc-root-scope-guard-eval` follow-up PR #595 proposed. This is also
+  the control-flow-heavy path #331 itself named as the prototype target — 11
```
pmatos: **Minor: off-by-one in the removed-call-site count.**

This says the migration collapses **11** `gc_unroot_frame` call sites, but the diff removes exactly **10** (`git diff main...HEAD -- src/interpreter/eval.rs | grep -c '^-.*gc_unroot_frame'` → 10). The PR description repeats the same '11'. Already corrected locally (not yet pushed):

```suggestion
  the control-flow-heavy path #331 itself named as the prototype target — 10
```

src/interpreter/eval.rs
pmatos: Correction — verified this is not dead code. `Generator.prototype.next` (builtins/iterators.rs ~4110-4124) does dispatch to `generator_next_state_machine` for `StateMachineGenerator` objects, but that path bridges back into this exact `Expression::Yield` delegate branch for per-item replay: `async_generator_next_state_machine_impl` (eval/generator_runtime.rs ~4178-4187) sets `self.generator_context = Some(GeneratorContext { .. })` whenever `inline_yield_target` is set (the `is_inline_replay` path), then calls `exec_state_machine_body`, which reaches `eval_expr`/`Expression::Yield` the same way the legacy `generator_next` does. The sync generator path has the equivalent bridge.

Confirmed empirically on this branch's release binary — both sync and async `yield*` delegating to a real (non-array) generator produce exactly one real advance of the inner generator per outer `.next()` call, matching Node byte-for-byte:

```js
function* inner() { console.log("a"); yield 1; console.log("b"); yield 2; }
function* outer() { yield* inner(); }
const g = outer();
g.next(); g.next(); g.next();
// jsse and Node both print "a" then "b" exactly once each, in step with the two yields
```

So this code path is live for real generators, not just array/iterable literals — the refactor is not touching dead code. Apologies for the earlier incorrect finding; that was posted by a sub-agent operating outside its assigned scope (it was dispatched only to review the `with_gc_root_scope` refactor's efficiency, not to post PR comments) without the reviewing coordinator's verification pass. The rest of the review (the "11 vs 10" `gc_unroot_frame` count finding on this same PR) stands and is accurate.

docs/adr/2026-09-10-2014-gc-root-scope-guard.md
```diff
@@ -0,0 +1,95 @@
...
+## Decision
+
+`Interpreter::with_gc_root_scope(|interp| …)` — a closure combinator that
+captures the current temp-root depth, runs `body`, and bulk-unroots on every
+exit path the closure takes (tail, early `return`, `?`) — is the seam for
+whole-body, single-frame native temp-root scoping. It is not a Drop-guard:
+`gc_temp_roots` stays a plain `Vec<u64>`, so there is no interior-mutability
+borrow tax (`RefCell`) on the GC hot path (`gc_root_value`, called from every
+allocation-adjacent site in the interpreter). The raw `gc_root_frame`/
+`gc_unroot_frame` primitive is retained for frames that genuinely nest or
```
pmatos: Altitude nit: this line's claim — that nested/interleaved frames are "the one case the combinator structurally cannot express, since it owns exactly one depth marker" — reads broader than it needs to. `with_gc_root_scope` calls nest fine by ordinary LIFO call-stack discipline (each invocation owns its own local `frame` marker on the same `Vec<u64>`); there's no structural reason two of them can't be nested.

Concretely, `destructure_array_assignment` (eval.rs:4288-4302 outer manual iterator root, plus a per-`Spread`-element inner `gc_root_frame`/`gc_unroot_frame` pair at eval.rs:4340) is exactly the "outer whole-body root + inner per-iteration frame" shape that an outer `with_gc_root_scope` wrapping the loop, plus an inner `with_gc_root_scope` per spread element, would express correctly — no `Rc<RefCell<...>>` needed for that inner frame specifically.

Not asking for that migration in this PR (small mechanical slice, reasonable to scope out), just flagging so the stated rationale doesn't steer a future reader away from an otherwise-safe simplification of the inner frame at eval.rs:4340.

src/interpreter/eval.rs
pmatos: Follow-up: my "correction" above was itself overstated in the same direction — sorry for the back-and-forth, posting the fuller picture now.

Traced `generator_next_state_machine` (eval/generator_runtime.rs ~867-1150): the `StateTerminator::Yield { is_delegate: true, .. }` arm is a **complete, separate, correctly-suspending** yield* implementation. It calls `get_iterator`/`next` itself and, on a non-done result, suspends by storing `delegated_iterator: Some(DelegatedIteratorInfo { iterator, next_method, resume_state, .. })` on the object and returning the raw iterator result directly (spec §14.4.14) — no replay, no `self.generator_context`, no call into `eval_expr`/`Expression::Yield`. Since `transform_generator` runs unconditionally (no bail) and `transform_yielding_expression` (generator_transform.rs ~761) always lowers a `Yield` node into this terminator, this native path is what a plain `yield* expr;` actually runs through — confirming the original finding's substance, not my correction.

There *is* a second, documented mechanism: a comment at generator_runtime.rs:4301-4304 says "Any `Completion::Yield` from `exec_statements` is an inline yield: it came from a loop body or complex control flow that isn't decomposed by the state machine transformer" — and that path does reuse `self.generator_context`/`SentValueBindingKind::InlineYield`, which *would* reach this PR's code. But I could not get a repro to exercise it for `yield*` specifically — a bare statement, `yield*` inside a `for` loop, and `yield*` inside `try/finally` all produced zero replayed side effects (checked via a side-effect counter + console.log, matching Node exactly each time), consistent with all three still going through the native `delegated_iterator` path above.

Net: I can't produce a case where this PR's refactored code executes at all for `yield*`; if such a case exists it's a narrower "complex control flow the transform can't decompose" edge case, not "every real generator" and not what my correction claimed either. The GC-rooting fix is still correct in isolation, just possibly for a much smaller/unconfirmed surface than either of us stated. Apologies again for the noise on this thread.

pmatos: Coordinator here, with the final picture after independently verifying this with direct instrumentation (a temporary `eprintln!` inside the exact `Expression::Yield` delegate branch, not part of the diff).

**Confirmed, more strongly than the black-box check above:** across 12 constructs — bare `yield*`, `while`/`do-while`/nested-`while` loops, `try/finally`, `switch`, `for-in`, labeled `continue`, `yield*` as a binary/call/ternary/array-literal operand, and an async generator — the instrumented line was never hit. Every one of those went through `generator_runtime.rs`'s `StateTerminator::Yield { is_delegate: true, .. }` arm instead, which suspends via `delegated_iterator` directly.

So: not "dead code for every real generator" (too strong — the `InlineYield` fallback at `generator_runtime.rs:4301-4304` is real, documented code that would reach this branch for control flow the transform can't decompose), but also not confirmed live for any construct anyone here could build. Filed **jsse#625** to track the open question (is `InlineYield` ever actually triggered, for `yield*` or plain `yield`, and should `self.generator_context`'s replay path be deleted) and to flag that `CLAUDE.md`'s "replay-based" generator description is likely stale.

Updated the ADR and PR description to state this plainly rather than the original "concrete... control-flow-heavy path" framing, which overclaimed current relevance. The `with_gc_root_scope` migration itself stands — it's a correct, fully-tested simplification of whatever executes this loop, if anything does; its practical value pending jsse#625.

docs/adr/2026-09-10-2014-gc-root-scope-guard.md
pmatos: Fixed (already applied in df45887, before you'd caught it here — same count). Thanks for the catch.

docs/adr/2026-09-10-2014-gc-root-scope-guard.md
pmatos: Fixed in 5e3f5eb — reworded to say plain LIFO nesting composes fine, and the raw primitive is reserved specifically for the interleaved/overlapping-lifetime shape (inner frame must truncate independently while an outer one stays open across repeated early exits), not nesting in general. Agreed this was overstated; not pulling destructure_array_assignment into this PR's scope, but the ADR no longer implies nesting itself is the blocker for it.

PR #626, opened 2026-09-10
refactor(typedarray): one receiver guard for %TypedArray% getters + validators

pmatos:
## Problem

The `%TypedArray%.prototype` module was **shallow at its receiver-validation seam**. The four
prototype getters — `byteOffset`, `byteLength`, `length`, `buffer` — each hand-spelled the same
receiver dance: `as_object_id → get_object → borrow → typed_array_info → (detach/OOB branch) → else
throw "not a TypedArray"`. The `validate_typed_array` seam already hides exactly that prologue, but
only under a **throw-on-detached** policy, so the getters (which must **return 0** on a detached or
out-of-bounds receiver, per spec `TypedArrayLength`) could not use it and re-opened the whole dance a
fourth way. Beside them sat a three-member validator sibling family
(`validate_typed_array` / `validate_uint8array` / `validate_uint8array_no_detach_check`) differing
only in brand string and detach policy. The interface (`this`) was trivial; the implementation
(brand + borrow + policy branch + two spellings of the throw) was repeated verbatim — the definition
of a shallow module, and drift had already started (`"not a TypedArray"` vs `"not a Uint8Array"` vs
`"typed array is detached"`).

## What changed (in `codebase-design` terms)

Two file-private **seams** in `typedarray.rs`, **extending what the existing receiver guard hides**
rather than adding a fourth parallel copy — the distinction that makes this a **deepening**, not
straggler-adoption of `validate_typed_array`:

- `with_typed_array_ref<R>(this, brand_msg, f)` — the shared brand-check prologue: resolve the
  receiver, run a **pure** `f` against the borrowed `TypedArrayInfo`, else throw a caller-chosen brand
  message (the object borrow drops before the error is built).
- `ta_number_getter(this, payload)` — the numeric-getter combinator that bakes in the brand-throw, the
  spec `TypedArrayLength → 0`-on-detached-or-OOB collapse (so `payload` is reached **only** on a live
  view), and the `Normal(number)` wrapping.

**Leverage**: the "what is a valid TypedArray receiver, and what happens when it is detached" decision
now lives once. **Locality**: a future resizable-buffer OOB fix or brand-message correction is a
one-line change behind the seam, not a 7-site sweep. **Test surface**: the drift-prone rule is now
directly unit-testable — assert `0` on a detached view *and that the payload is never invoked*.

**Before** (solid = interface, dashed = inside the implementation):

```mermaid
graph LR
  G1[byteOffset] -.-> P[brand-check + borrow + detach/OOB branch + throw-tail]
  G2[byteLength] -.-> P
  G3[length] -.-> P
  G4[buffer] -.-> P
  V1[validate_typed_array] -.-> P
  V2[validate_uint8array] -.-> P
  V3[validate_uint8array_no_detach_check] -.-> P
```

**After**:

```mermaid
graph LR
  G1[byteOffset] --> N[ta_number_getter]
  G2[byteLength] --> N
  G3[length] --> N
  N --> R[with_typed_array_ref]
  G4[buffer] --> R
  V1[validate_typed_array] --> R
  V2[validate_uint8array] --> R
  V3[validate_uint8array_no_detach_check] --> R
  R -.-> P[resolve + brand-throw + snapshot]
```

The 3 numeric getters collapse to one-line payload closures (zero-clone, read through the borrow);
`buffer` uses the kernel directly; the 3 validators layer their brand/kind + `check_detached*`
refinements on the kernel. `validate_typed_array`'s signature and its 19 callers are untouched.

## Behaviour-preserving

Test-first (a red test module pinning the interface before it existed). Exact error strings, the
detached-only vs detached-or-OOB distinction per site, the getters' return-0 semantics, and `buffer`'s
`buffer_object_id: None → "not a TypedArray"` fall-through are all preserved.

> **Follow-up (not fixed here):** `buffer` throwing `"not a TypedArray"` when a genuine TypedArray has
> an unmaterialised buffer-object wrapper (`buffer_object_id: None`) is a pre-existing spec oddity, kept
> as-is to stay behaviour-preserving. Worth a separate issue.

## Quality gate

- `cargo test --lib`: 675 passing + **5 new** (`ta_number_getter_tests`); 0 failures.
- `./scripts/lint.sh`: rustfmt + clippy (`-D warnings`, incl. perf-counters) clean.
- test262 `built-ins/TypedArray` + `built-ins/Uint8Array`: **3016/3016**; the 4 getters `built-ins/TypedArray/prototype/{byteOffset,byteLength,length,buffer}`: **128/128**. **0 regressions.**

Diff: **1 file** (`typedarray.rs`) + a `CONTEXT.md` term — matches the blast-radius-1 score.

## Selection (auditable)

Full review: [`.architecture/reviews/2026-09-11-typedarray-getter-receiver-guard.md`](.architecture/reviews/2026-09-11-typedarray-getter-receiver-guard.md).

- **Pick**: `typedarray-getter-receiver-guard` — **22/25** (leverage 4, locality 4, blast radius 1, heat 5).
- **Runner-up candidate**: `gc-root-scope-guard-eval` — also **22/25** (blast radius 3). The top two
  **tied**; the deterministic lower-blast-radius tie-break decided it. `gc-root-scope-guard-eval` is the
  natural next firing.
- **Winning design**: getter-first combinator (`with_typed_array_ref` + `ta_number_getter`), zero new
  named types.
- **Runner-up design**: a fully policy-parameterized `guard_ta_receiver` with four enums
  (`Brand`/`DetachPolicy`/`Scope`/`Guard`). It **lost** on depth (four types to learn dilutes
  behaviour-per-unit-of-interface) and seam placement (its enum matrix seams policy combinations with
  zero or one user — hypothetical seams by the "two adapters = real seam" test), despite a marginally
  richer test surface.

## CONTEXT.md

Adds the **Receiver Guard** term — the receiver-validation prologue family that `validate_typed_array`,
`require_array_buffer` (#570), and now `with_typed_array_ref` all instance.

## Proposed ADR

> **ADR: Receiver guards are combinators, not policy-parameterized functions, when a policy varies by
> caller group.** When a receiver-validation seam must serve caller groups with genuinely different
> post-validation policies (e.g. TypedArray getters return 0 on a detached view where prototype methods
> throw), prefer a small combinator that bakes the shared group's rule in
> (`ta_number_getter`) layered over a policy-free kernel (`with_typed_array_ref`), over one function
> parameterized by a policy enum. Rationale: the combinator keeps the hot caller's interface to one
> function and one closure, seams only rules that have two or more real adapters, and makes the
> group's rule structurally unforgettable — where an enum surfaces every policy combination, including
> single-user ones, at every call site. Supersedes nothing; complements the existing receiver-guard
> family (#543, #570).

---
🤖 Generated with Claude Code

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -67,3 +67,18 @@ _Avoid_: permanent root, closure root.
 **Rooted Slot**:
 A GC-traced container an anchor pins once and the owner then mutates in place, for a capture whose value is *replaced* over the anchor's lifetime. A native closure's own `Rc<RefCell<…>>` state is invisible to the tracer, and re-pinning each replacement would retain every superseded value, so the slot — not the value — is what gets pinned. `RootedPair` in `builtins/iterators.rs` is the two-slot case: the iterator an iterator helper is currently drawing from, plus that iterator's `next` method.
 _Avoid_: root cell, traced box, rooted buffer.
+
+## Builtins
+
+**Receiver Guard**:
+The single prologue a native method routes its `this` through to brand-check the
+receiver and hand back an owned snapshot of its kind-specific info — or throw the
+brand `TypeError` — so callers never re-spell the `as_object_id → get_object →
+borrow → <kind>_info` dance. The family: `validate_typed_array` /
+`with_typed_array_ref` (any TypedArray), `require_array_buffer` /
+`require_shared_array_buffer` (`builtins/typedarray.rs`). A guard may be
+parameterized by a detach policy: `ta_number_getter` layers the numeric getters'
+spec `TypedArrayLength → 0`-on-detached-or-out-of-bounds rule on top of the guard,
+where `validate_typed_array` instead throws. The snapshot is owned so the object
+borrow drops before any `create_type_error` (which mutates the object arena).
+_Avoid_: receiver check, brand check (for the whole prologue), this-unwrap.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -72,13 +72,17 @@ _Avoid_: root cell, traced box, rooted buffer.
 
 **Receiver Guard**:
 The single prologue a native method routes its `this` through to brand-check the
-receiver and hand back an owned snapshot of its kind-specific info — or throw the
-brand `TypeError` — so callers never re-spell the `as_object_id → get_object →
-borrow → <kind>_info` dance. The family: `validate_typed_array` /
-`with_typed_array_ref` (any TypedArray), `require_array_buffer` /
-`require_shared_array_buffer` (`builtins/typedarray.rs`). A guard may be
-parameterized by a detach policy: `ta_number_getter` layers the numeric getters'
-spec `TypedArrayLength → 0`-on-detached-or-out-of-bounds rule on top of the guard,
-where `validate_typed_array` instead throws. The snapshot is owned so the object
-borrow drops before any `create_type_error` (which mutates the object arena).
+receiver — or throw the brand `TypeError` — so callers never re-spell the
+`as_object_id → get_object → borrow → <kind>_info` dance. The family:
+`require_array_buffer` / `require_shared_array_buffer` (`builtins/typedarray.rs`)
+hand back an owned snapshot of the buffer's kind-specific info; `with_typed_array_ref`
+is the leaner kernel form — it runs a caller-supplied closure against the borrowed
+`TypedArrayInfo` and returns whatever the closure returns (a scalar, an owned
+snapshot via `TypedArrayInfo::clone`, or anything else), so ownership is the
+caller's choice, not the guard's. `validate_typed_array` layers a snapshot +
+detached/OOB throw on top of the kernel. A guard may be parameterized by a
+detach policy: `ta_number_getter` layers the numeric getters' spec
+`TypedArrayLength → 0`-on-detached-or-out-of-bounds rule on top of the kernel,
+where `validate_typed_array` instead throws. Either way, any borrow the guard
+holds drops before `create_type_error` runs (it mutates the object arena).
 _Avoid_: receiver check, brand check (for the whole prologue), this-unwrap.
--- a/src/interpreter/builtins/typedarray.rs
+++ b/src/interpreter/builtins/typedarray.rs
@@ -5645,8 +5645,7 @@ fn ta_buffer_getter(interp: &mut Interpreter, this_val: &JsValue) -> Completion 
         ta.buffer_object_id
     }) {
         Ok(Some(buf_id)) => Completion::Normal(JsValue::object(buf_id)),
-        Ok(None) => Completion::Throw(interp.create_type_error("not a TypedArray")),
-        Err(throw) => throw,
+        Ok(None) | Err(_) => Completion::Throw(interp.create_type_error("not a TypedArray")),
     }
 }
 
@@ -5662,7 +5661,7 @@ fn ta_number_getter(
     payload: impl FnOnce(&TypedArrayInfo) -> f64,
 ) -> Completion {
     match with_typed_array_ref(interp, this_val, "not a TypedArray", |ta| {
-        if ta.is_detached.get() || is_typed_array_out_of_bounds(ta) {
+        if is_ta_detached_or_oob(ta) {
             0.0
         } else {
             payload(ta)
@@ -5695,13 +5694,10 @@ fn validate_uint8array_no_detach_check(
     interp: &mut Interpreter,
     this_val: &JsValue,
 ) -> Result<TypedArrayInfo, Completion> {
-    let ta = with_typed_array_ref(interp, this_val, "not a Uint8Array", TypedArrayInfo::clone)?;
-    if !matches!(ta.kind, TypedArrayKind::Uint8) {
-        return Err(Completion::Throw(
-            interp.create_type_error("not a Uint8Array"),
-        ));
-    }
-    Ok(ta)
+    let ta = with_typed_array_ref(interp, this_val, "not a Uint8Array", |ta| {
+        matches!(ta.kind, TypedArrayKind::Uint8).then(|| ta.clone())
+    })?;
+    ta.ok_or_else(|| Completion::Throw(interp.create_type_error("not a Uint8Array")))
 }
 
 fn check_detached(interp: &mut Interpreter, ta: &TypedArrayInfo) -> Result<(), Completion> {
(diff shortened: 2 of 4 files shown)


src/interpreter/builtins/typedarray.rs
```diff
@@ -5652,71 +5622,88 @@ fn require_shared_array_buffer(
...
+/// door the prototype getters and the `validate_*` helpers route `this` through.
+fn with_typed_array_ref<R>(
     interp: &mut Interpreter,
     this_val: &JsValue,
-) -> Result<TypedArrayInfo, Completion> {
+    brand_msg: &str,
+    f: impl FnOnce(&TypedArrayInfo) -> R,
+) -> Result<R, Completion> {
     if let Some(o) = this_val.as_object_id()
         && let Some(obj) = interp.get_object(o)
```
pmatos: **Efficiency**: `with_typed_array_ref` resolves the receiver via `interp.get_object(o)` (clones an `ObjectHandle` — a refcounted handle clone+drop) instead of the cheaper `interp.get_object_cell(o)` (borrows a `&ObjectHandle`, no clone). Before this PR, the `byteOffset`/`byteLength`/`length` getters called `get_object_cell` directly — this refactor regresses all three onto the more expensive path on what is close to the hottest TypedArray property-read.

The swap is safe: the borrow only needs to live for the duration of the `f` callback, and `interp.create_type_error(brand_msg)` is only reached after that scope has ended (same shape the deleted getter bodies compiled with). Suggested fix:
```rust
    if let Some(o) = this_val.as_object_id()
        && let Some(obj) = interp.get_object_cell(o)
    {
```

Applied as a fix in this review pass; `cargo test --lib` (675/675) and the `built-ins/TypedArray` + `built-ins/Uint8Array` test262 subset (3016/3016) both stay green after the change.

src/interpreter/builtins/typedarray.rs
```diff
@@ -5652,71 +5622,88 @@ fn require_shared_array_buffer(
...
+fn validate_typed_array(
+    interp: &mut Interpreter,
+    this_val: &JsValue,
+) -> Result<TypedArrayInfo, Completion> {
+    let ta = with_typed_array_ref(interp, this_val, "not a TypedArray", TypedArrayInfo::clone)?;
+    check_detached_or_out_of_bounds(interp, &ta)?;
+    Ok(ta)
 }
 
 fn validate_uint8array(
```
pmatos, line 5684: **Simplification**: `validate_uint8array` and `validate_uint8array_no_detach_check` duplicate an identical 5-line "not a Uint8Array" kind-check block verbatim — exactly the kind of duplication `with_typed_array_ref` was extracted to stop, and by this PR's own "two adapters = real seam" bar (from the Design C verdict in the PR description) these two functions qualify. `validate_uint8array` can delegate instead:
```rust
fn validate_uint8array(
    interp: &mut Interpreter,
    this_val: &JsValue,
) -> Result<TypedArrayInfo, Completion> {
    let ta = validate_uint8array_no_detach_check(interp, this_val)?;
    check_detached(interp, &ta)?;
    Ok(ta)
}
```
Preserves the exact error strings and check order (brand -> kind -> detach) for every case. Applied as a fix in this review pass; tests and the TypedArray/Uint8Array test262 subset stay green.

src/interpreter/builtins/typedarray.rs
```diff
@@ -1311,64 +1311,34 @@ impl Interpreter {
...
-                {
-                    return Completion::Normal(JsValue::object(buf_id));
-                }
-            }
-            Completion::Throw(interp.create_type_error("not a TypedArray"))
-        });
+        // buffer getter (returns the ArrayBuffer object; no detached/OOB check —
+        // the buffer survives detachment. `buffer_object_id: None` on the rare
+        // wrapperless construction path preserves the "not a TypedArray" throw.)
+        self.define_getter(
```
pmatos: **Simplification** (minor): the `buffer` getter's body is a 3-arm `match` written inline in the `define_getter` closure, breaking the one-liner pattern the other 3 getters get via `ta_number_getter`. Extracting a small `ta_buffer_getter(interp, this_val) -> Completion` (mirroring `ta_number_getter`'s shape) keeps the getter-registration block uniform:
```rust
self.define_getter(proto_id, "buffer", |interp, this_val, _args| {
    ta_buffer_getter(interp, this_val)
});
```
Applied as a fix in this review pass (low severity on its own, but cheap and keeps the 4 getters visually consistent).

CONTEXT.md
```diff
@@ -67,3 +67,18 @@ _Avoid_: permanent root, closure root.
 **Rooted Slot**:
 A GC-traced container an anchor pins once and the owner then mutates in place, for a capture whose value is *replaced* over the anchor's lifetime. A native closure's own `Rc<RefCell<…>>` state is invisible to the tracer, and re-pinning each replacement would retain every superseded value, so the slot — not the value — is what gets pinned. `RootedPair` in `builtins/iterators.rs` is the two-slot case: the iterator an iterator helper is currently drawing from, plus that iterator's `next` method.
 _Avoid_: root cell, traced box, rooted buffer.
+
+## Builtins
+
+**Receiver Guard**:
+The single prologue a native method routes its `this` through to brand-check the
+receiver and hand back an owned snapshot of its kind-specific info — or throw the
```
pmatos: **Doc-accuracy** (minor): this glossary entry says the Receiver Guard prologue "hand[s] back an owned snapshot of its kind-specific info", and lists `with_typed_array_ref` in the same family as `require_array_buffer`/`require_shared_array_buffer`. But `with_typed_array_ref` is generic (`Result<R, Completion>`) — its `ta_number_getter` caller instantiates `R = f64` and the `buffer` getter instantiates `R = Option<u64>`, neither of which is "an owned snapshot of kind-specific info". Only `validate_typed_array`/`validate_uint8array*` (which pass `TypedArrayInfo::clone` as the closure) produce that snapshot; `require_array_buffer`/`require_shared_array_buffer` are accurately described but `with_typed_array_ref` itself is not. Reworded in this review pass to describe `with_typed_array_ref` as the leaner kernel form whose return type is the caller's choice.

src/interpreter/builtins/typedarray.rs
```diff
@@ -5652,71 +5622,88 @@ fn require_shared_array_buffer(
...
+/// "detached-or-out-of-bounds ⇒ 0" collapse (so `payload` is reached only on a
+/// live view), and the `Completion::Normal(number)` wrapping. Shared by the
+/// `byteOffset` / `byteLength` / `length` getters.
+fn ta_number_getter(
+    interp: &mut Interpreter,
+    this_val: &JsValue,
+    payload: impl FnOnce(&TypedArrayInfo) -> f64,
+) -> Completion {
+    match with_typed_array_ref(interp, this_val, "not a TypedArray", |ta| {
+        if ta.is_detached.get() || is_typed_array_out_of_bounds(ta) {
```
pmatos: **Simplification / missed opportunity** (low severity, informational only — not applied as a fix): `ta_number_getter` inlines `ta.is_detached.get() || is_typed_array_out_of_bounds(ta)` rather than calling a shared predicate, even though `check_detached_or_out_of_bounds` (a few lines below) wraps the identical condition for the throw case. This exact inline expression already existed twice before this PR (lines ~1434 and ~1561, both predating this PR per `git blame`), so this is a pre-existing local pattern the PR continues rather than new duplication it introduces. Worth a follow-up `is_detached_or_out_of_bounds(&TypedArrayInfo) -> bool` predicate shared by all 4 sites, but that touches code outside this diff's scope so left unfixed here.

pmatos:
**Conventions note** (informational, no code fix): `CLAUDE.md` line 66 states "After any implementation work, run the full test262 suite." The PR's "Quality gate" section reports `cargo test --lib` plus test262 restricted to `built-ins/TypedArray` + `built-ins/Uint8Array` (3016/3016) and the 4 getter directories (128/128) — a targeted subset, not the full suite via `uv run python scripts/run-test262.py` with no path argument.

Caveat: `.github/workflows/ci.yml` itself also only runs a smoke set + a 10%-sample test262 run per PR; the full suite runs nightly only (`nightly-test262-coverage.yml`). So this likely reflects existing repo-wide PR practice rather than a lapse specific to this PR — flagging for visibility, not as a blocker.

PR #657, opened 2026-09-20
fix(generators): scope pending iterator-close list to each generator activation

### LINKED ISSUE #650: bug: JetStream lazy-collections fails with "TypeError: Generator is already running"
## Description

The JetStream `lazy-collections` workload fails immediately with:

```
TypeError: Generator is already running
```

It exits with code 1 after ~0.07 s, so it fails at or near start-up. Same result with `--bytecode`. It also failed on the 2026-03-17 build, so this is not a regression. The workload (`generators/lazy-collections.js`) builds lazy collections out of nested generators, so the guard that raises this error may be tripping on a legitimate re-entry (unverified; jsse generators are replay-based, see the architecture notes). Related area: #53.

## Reproduction

```bash
# JetStream, pinned to the revision used in docs/perf reports; assets ship compressed
git clone --depth 1 https://github.com/WebKit/JetStream /tmp/JetStream && cd /tmp/JetStream
git fetch --depth 1 origin c603c04db8505477867974a69789309ded2cc948 && git checkout c603c04db8505477867974a69789309ded2cc948
npm install && npm run decompress -- --quiet     # needs node >= 22
# from the jsse checkout
cargo build --release
uv run python scripts/run-jetstream.py --test lazy-collections --iterations 1 --timeout 120 --engine target/release/jsse --jetstream /tmp/JetStream
```

Needs minimizing to a small generator test case; if it reproduces without JetStream, a test262-extra test should go with the fix.

## Environment

jsse v0.8.1 (`083746f`, release build), buildbox3 (AMD EPYC 7501, Linux 6.1), JetStream `c603c04`, `--iterations 1 --timeout 120 -j 1`, N=3 repeats. Found while collecting data for the 2026-09-18 comparison against the 2026-03-17 build (`7a4d095`).

pmatos:
## Summary

`Interpreter::pending_iter_close` is one interpreter-wide list of the iterators of every open generator `for-of`. Each generator/async-generator driver drained the whole list when it yielded and parked it in `generator_inline_iters[<that generator>]`. So when generator A was mid-`for-of` and its body ran generator B to a `yield`, B stole A's open iterators. A later `B.return()` (an early `return`/`break` out of a `for-of` over B, as `slice`/`take`/`every` do in `lazy-collections.js`) then called `.return()` on iterators B never owned:

- A's own for-of source, still executing → `TypeError: Generator is already running` (the #650 failure);
- a suspended generator → silently closed early, so A's loop ended prematurely with wrong results and no error.

Fix: each generator/async-generator activation gets a base index (`iter_close_base`) into `pending_iter_close`. The drivers only take entries at `base..` on yield, and truncate back to the base when the activation returns, so enclosing activations' entries stay put (and stay GC-rooted). The take helper does not allocate when nothing is pending.

Spec basis: GeneratorValidate (only the receiver's own state), GeneratorResumeAbrupt (only the receiver's execution context is unwound), ForIn/OfBodyEvaluation (a loop's iterator is closed by its own generator), and the async analogues.

## Tests

- New `test262-extra/generator-nested-activation-does-not-steal-for-of-iterators.js` and `async-generator-nested-activation-does-not-steal-for-await-iterators.js` (both fail on the unfixed binary, pass fixed): nested `return()`/`next()`, the lazy-collections `map`/`chunk`/`take` pipeline, early `break` over a nested generator, outer `return()` closing its own iterator exactly once, and `return()` on a nested generator that is itself suspended in a `for-of` (closes only its own iterator).
- Unit test `generator_activations_leave_no_pending_iterators_behind` (`pending_iter_close` empty, `iter_close_base == 0` after nested/throwing generators).
- The plan's check on `generator_return/throw_state_machine` (and async counterparts) running outside the scope is covered by the last case above; no wrapper needed there.

## Verification

- `./scripts/lint.sh`: OK. `cargo test --release`: all pass.
- Full `run-test262.py -j 32`: 99911/99911, `regressions: 0` (baseline `origin/main:test262-pass.txt`).
- `test262-extra/`: 332/332 in default and `--bytecode`. `run-custom-tests.py`: 11/11. `run-library-tests.sh acorn`: green (13507, cross-checked vs Node).
- JetStream `generators/lazy-collections.js` (pinned `c603c04`) plus `new Benchmark().runIteration()`/`validate`: unfixed binary throws `TypeError: Generator is already running` (default and `--bytecode`); fixed binary completes and validates in both. `scripts/run-jetstream.py` itself refused to run (busy-host idle gate, because the full test262 run was in flight), so the workload was driven directly rather than through the runner.

## Known, scoped out

The whole-list `retain`/`any` operations in the for-of head/exit code and `unroot_for_of_iterator` still scan below `iter_close_base`. This is only observable if an enclosing and a nested activation iterate the *same* iterator object; no failing case exists, so it is left as a follow-up rather than widening this diff.

Closes #650


pmatos:
Automated code-review pass pushed 82ac045: scoped the pending-iterator scans to the running activation's window, dropped the unconsumed for-of push in `async_function_resume`, and folded the duplicated save-at-yield blocks into `stash_pending_iter_close` / `with_iter_close_scope`. Two new test262-extra tests cover shared iterators across nested activations and an async function started inside a generator. Full test262 99,911/99,911, test262-extra 336/336, cargo test and lint green.

Not fixed (needs work outside this diff): with the scoped list, an async generator's `yield*` return path (`yield_star_return_after_unwrap`) no longer has the global steal masking that it doesn't unwind its own `for_of_stack`. For `for await (x of L) { yield* inner(); }` closed via `return()`, jsse now logs `ret:I` only; node logs `ret:I,ret:L` (base commit logged both, in the wrong order). Also pre-existing and untouched: async `yield*` return skips the outer `finally`, and `yield*` error paths skip enclosing loops.

pmatos:
Autofix pass: `workflow-lint` was the only failing check, and it was unrelated to this PR's generator changes (the branch touches no `.github/` files).

**Cause:** dtolnay moved the rolling `v1` tag of `dtolnay/rust-toolchain` from `6c977a6` to `02cb101` (six commits ahead). zizmor 1.26.1's online `ref-version-mismatch` audit then flagged the seven `# v1` pin comments as stale, so `workflow-lint` fails on any branch that runs it after the tag moved (last green `main` run predates the move).

**Fix (`b7e47fb`):** bumped all seven pins (`ci`, `release`, `nightly-test262-coverage`, `miri`, `codeql`, and `fuzz` twice) to `02cb101` with the `# v1` comment. Verified locally with `uvx zizmor@1.26.1 .github/workflows/`: no findings. The upstream diff between the two commits is limited to `action.yml` (`--force-non-host`), `scripts/update-revs.sh`, and the action's own CI.

**Alternative considered:** suppressing `ref-version-mismatch` for this action in `.github/zizmor.yml`. That would stop the treadmill (the `v1` tag will move again), but it is a lint-policy change that doesn't belong in a generator bugfix PR. Worth a separate issue if the churn recurs.

PR #658, opened 2026-09-20
fix(generators): decompose class heritage/computed keys to close InlineYield gap

### LINKED ISSUE #625: Is the legacy self.generator_context/exec_statements yield* path (InlineYield fallback) still reachable?
## Context

While reviewing PR #624 (migrating `eval.rs`'s `yield*` delegation loop to
`with_gc_root_scope`), a code-review pass questioned whether that loop —
which reads `self.generator_context` and replays via `eval_expr`'s
`Expression::Yield` arm — is reachable by any real JS program.

Independently confirmed by direct instrumentation (a temporary `eprintln!`
inside the exact branch, removed before merging): **zero hits** across all
of the following, each checked on both this branch's release binary:

- a bare `yield* iterable;` statement
- `yield*` inside a `while` loop, a `do...while` loop, and nested `while`
  loops
- `yield*` inside `try { } finally { }`
- `yield*` inside a labeled loop with `continue label`
- `yield*` inside a `switch` case
- `yield*` inside a `for...in` loop
- `yield*` as an operand of a binary expression, a call argument, a ternary
  branch, and an array-literal element
- `async function*` with a bare `yield*` (the async-generator equivalent
  branch)

Every one of these instead went through `generator_runtime.rs`'s
`generator_next_state_machine`/`async_generator_next_state_machine_impl`,
specifically the `StateTerminator::Yield { is_delegate: true, .. }` arm,
which suspends by storing `delegated_iterator: Some(DelegatedIteratorInfo
{ .. })` directly on the generator object (spec §14.4.14) — no replay, no
call back into `eval_expr`.

## The `InlineYield` fallback does exist, just not (yet) triggered here

`generator_runtime.rs:4301-4304` documents a second mechanism:

> Any `Completion::Yield` from `exec_statements` is an inline yield: it came
> from a loop body or complex control flow that isn't decomposed by the
> state machine transformer. Use `InlineYield` to re-enter the same state
> and fast-forward past previous yields.

This is real code (`SentValueBindingKind::InlineYield`, `generator_context`
set at `generator_runtime.rs:71/237/372/749/4185/6133`), and it's what would
route through `eval.rs`'s `Expression::Yield` arm (both the `yield*` branch
PR #624 touched and the plain-`yield` branch it didn't). But `transform_generator`/
`transform_async_generator` run unconditionally with no bail/error return
(`generator_transform.rs:348`), and nothing found in this investigation
triggers the fallback for `yield*` specifically.

## Open questions

1. Is there a real, reachable JS construct — for `yield*` or for plain
   `yield` — that still exercises the `InlineYield`/`self.generator_context`
   replay path today? If none exists, is the mechanism dead code that should
   be deleted (and `self.generator_context`, `current_yield`/`target_yield`,
   `GeneratorResumeKind`, and the `Expression::Yield` arm's manual state
   along with it), or is it deliberately retained for some case the state-
   machine transform is not expected to ever fully cover?
2. `CLAUDE.md`'s Architecture Notes currently say "Generators use a
   replay-based approach (re-execute the function body, fast-forwarding past
   previous yields)." That description matches the `InlineYield` fallback,
   not the now-dominant `StateTerminator`/`delegated_iterator` state-machine
   path. Worth updating once (1) is resolved, so the doc reflects what
   actually executes for ordinary generators.

## Unrelated anomaly noticed in passing (not investigated further)

Wrapping a `yield*` delegation in a `while` loop that re-enters it on a
second iteration showed the inner iterable's `next()` called **two** extra
times at the loop-iteration boundary (a jump from N to N+2 calls instead of
N+1) compared to a plain `for`/`do-while`/labeled-loop/`switch` wrapping,
which all showed exactly one `next()` call per external `.next()`. Not
confirmed as a spec violation or characterized further — flagging in case it
recurs when (1) above is investigated. No reduced repro was minimized.

Refs #331 (this is the issue PR #624 traces back to), #595.

pmatos:
## What

Answers issue #625's first open question with a concrete, reachable trigger:
a `yield`/`yield*`/`await` inside a generator's own class declaration's
`extends` clause or a computed method/field key.

Per spec `ClassDefinitionEvaluation` (§15.7.14), a class's heritage expression
and each element's computed key evaluate in the *enclosing* scope, not inside
a nested function/class boundary. `generator_analysis.rs`'s three AST walkers
(`analyze_statement`/`analyze_expression`, `contains_yield`/`expr_contains_yield`,
`contains_suspension`/`expr_contains_suspension`) all stopped at
`Statement::ClassDeclaration`/`Expression::Class`, treating the class as an
opaque boundary — correct for method/field bodies, wrong for the class's own
heritage/computed-key expressions. That gap meant a generator body containing
one of these constructs was misclassified as yield-free, took the
`create_simple_machine` fast path (or, once detected, still wasn't decomposed
by the transform), and any embedded yield fell through to the legacy
`self.generator_context`/`SentValueBindingKind::InlineYield` replay mechanism
this issue set out to investigate — re-executing the whole class statement
(or worse, the whole function body) on every resume.

## Fix

- **Detection** (`generator_analysis.rs`): recurse into a class's `super_class`
  and each element's computed `PropertyKey`, exactly like the existing
  `Expression::Object` computed-key handling. Sync (`contains_yield`) and
  async (`contains_suspension`) mirrors, both in the statement and expression
  analysis functions.
- **Decomposition** (`generator_transform.rs`): add `Statement::ClassDeclaration`
  and `Expression::Class` arms that hoist suspending heritage/keys into temp
  vars via the existing `transform_yielding_expression(..., Some(Variable(tv)))`
  pattern — heritage first (source order), then each computed key in
  declaration order. A class with no suspending heritage/keys never enters
  this path (gated by the same detection functions), so plain classes —
  the overwhelming common case — see zero transform difference.
- **Docs**: reworded the `CLAUDE.md` Architecture Notes line that described
  the InlineYield fallback as if it were the primary generator mechanism —
  it isn't; the `StateTerminator`/state-machine path is. Answers the issue's
  second open question.

## Known, accepted limitation

When a class has *both* a heritage expression that throws (and itself
contains no suspension, so it's left un-hoisted) *and* a computed key that
does need hoisting, the hoisted key's side effect (and any yield inside it)
can now run **before** the heritage `TypeError` — spec order is the reverse
(heritage evaluated and checked before any element). Fully closing this
would require making `ClassDefinitionEvaluation` itself resumable/interleaved
with the state machine, not just hoisting independent sub-expressions —
out of scope here. Pinned with a dedicated regression test
(`generator-class-heritage-throw-vs-computed-key-side-effect-order.js`) so a
future change to this ordering is deliberate, not silent drift.

## Not addressed (left as-is, per issue scope)

- `self.generator_context`/`current_yield`/`target_yield`/`GeneratorResumeKind`/
  `InlineYield` and the `Expression::Yield` arm's manual replay state are
  **not** deleted — this investigation audited one AST-node family (class
  heritage/computed keys), not an exhaustive one, so the fallback is not
  proven dead code. Deleting defensive fallback machinery is a separate,
  larger decision this evidence doesn't support making yet.
- The issue's "unrelated anomaly" (`yield*` in a `while` loop showing an
  N→N+2 jump in delegate `next()` calls across a loop-iteration boundary)
  does not reproduce on this branch with a minimal repro and is not touched
  here — it stays exactly as open/unminimized as the issue described.
- Class field initializers: confirmed both jsse and Node reject
  `x = (yield 1);` as a field initializer with a parse-time `SyntaxError`
  despite the grammar's `[+Yield]` parameter threading, so they're not a
  reachable yield site and are untouched.
- `ClassStaticBlock` bodies: fixed `[~Yield, +Await, ~Return]` regardless of
  enclosing context, so a bare `yield` there is never a `YieldExpression` —
  not a trigger, untouched.

## Testing

- New unit tests in `generator_analysis.rs`/`generator_transform.rs` (TDD
  slices: detection, async mirror, decomposition, and a regression guard that
  a suspension-free class stays on the untouched fast path).
- New `test262-extra/` files, since test262 doesn't assert on side-effect/call
  counts and wouldn't have caught this bug or a regression:
  - `generator-class-computed-key-yield-single-evaluation.js`
  - `generator-class-heritage-yield-star-single-iterator-advance.js`
  - `generator-class-heritage-throw-vs-computed-key-side-effect-order.js`
    (known-limitation pin, see above)
  - `async-generator-class-computed-key-await-single-evaluation.js`
- `cargo test --release`: all pass.
- `uv run python scripts/run-test262.py -j 32` (full suite, unconditional
  per project convention): **99,911/99,911 scenarios pass, 0 regressions**
  against the `origin/main` baseline.
- `uv run python scripts/run-test262.py test262-extra/ -j 32`: 344/344 pass.
- `uv run python scripts/run-custom-tests.py`: 11/11 pass.
- `./scripts/lint.sh`: clean (rustfmt + clippy, both default and
  `perf-counters` feature).

Closes #625


pmatos:
Automated `/code-review --fix` pass (xhigh) pushed 189b186.

**Fixed**
- Class *expressions* were still opaque to `analyze_expression`, so `let C = class { [yield k]() {} }` had no yield points and replayed via InlineYield (ran side effects twice). Now analysed like declarations.
- Hoisting an anonymous class into a generated temp leaked `$return_0` / `$call_arg_*` as `.name`; now wrapped in `(0, class ...)`.
- New `ClassElement::computed_key()` + `class_scope_exprs()` replace five hand-rolled heritage/key walkers; `hoist_class_suspensions` clones and rewrites the key in place instead of re-listing fields.
- Corrected "enclosing scope" comments, fixed test262-extra feature name (`async-iteration`), replaced a weak unit test, added class-expression regression tests.

**Not fixed (design/scope limits, pre-existing or inherited from the existing hoisting arms)**
- Non-suspending heritage/keys evaluated before a later suspending key still run after it (ordering), and ToPropertyKey is deferred to class-definition time.
- Hoisted heritage/keys run outside classScope (named class expression TDZ not modelled).
- Flattened classes lose block/per-iteration `let`/`with` scoping (existing state-machine limits).
- `transform_for_in_statement` is a pre-existing stub emitting `Statement::Empty`; a for-in containing a suspending class now hits it. Needs a real for-in decomposition.

**Validation:** lint, `cargo build --release`, `cargo test --release`, full test262 (99,911/99,911, 0 regressions), test262-extra (348/348), custom tests (11/11).

pmatos:
Simplify pass (8c9aa94): consolidated the heritage/computed-key walkers into `ast::class_scope_exprs` (net −84 lines), added `ClassElement::key_mut`, shared one hoist helper, and replaced the `$..._N` name sniffing with a `generated_temps` set applied in `emit_expression_with_binding`. That also stops temp names leaking into anonymous functions/classes in conditional/sequence arms and fixes the false positive for a user variable named like `$a_1` (new test262-extra test). Skipped: hoisting all earlier keys once a later one suspends (behaviour change, already pinned by an existing test); avoiding one discarded key clone (compile-time only); inlining the one-line class_contains_* wrappers. Gate: lint, release build, cargo test, test262 (99,911/99,911, 0 regressions) and test262-extra (350/350) all green.

Issue #643, opened 2026-09-18
bug: delete via optional chain skips exotic-object [[Delete]] special cases (TypedArray/String/module namespace)


pmatos:
## Summary

`delete obj?.prop` (optional-chain delete) does not run the same `[[Delete]]` special-case handling as the plain `delete obj.prop` operator. Concretely, deleting a valid integer-indexed TypedArray element through an optional chain returns `true` (and never throws in strict mode) instead of returning `false` (and throwing `TypeError` in strict mode), because the optional-chain delete path is missing the Integer-Indexed exotic object, String exotic object, and module-namespace exotic `[[Delete]]` branches entirely.

## Spec

`[[Delete]]` internal method as invoked by the `delete` unary operator (ECMA-262 13.5.1.2 `UnaryExpression : delete UnaryExpression`), specifically:
- 10.4.5.4 `[[Delete]]` for Integer-Indexed exotic objects (TypedArrays) — must return `false` for a valid canonical numeric index.
- 10.4.6.2 `[[Delete]]` for String exotic objects.
- Module Namespace Exotic Objects `[[Delete]]`.

## Root cause

- `src/interpreter/eval.rs:824-942` (the plain `delete obj.prop` path) checks, in order: proxy → module-namespace exotic → TypedArray valid-integer-index exotic (return `false`, throw in strict mode) → String exotic index → ordinary configurable check.
- `src/interpreter/eval/access.rs:437-505` (`eval_delete_on_object`, reached only via optional-chain delete, `delete obj?.prop`, through `eval_delete_oc_tail`) has **only** the proxy check and the ordinary configurable check — no TypedArray, String-exotic, or module-namespace handling.

## Repro

```js
var ta = new Int8Array(1);
delete ta[0];                     // false (correct, both jsse and Node)

var obj = {ta: new Int8Array(1)};
delete obj?.ta[0];                // jsse: true -- should be false

(function(){ "use strict";
  var obj2 = {ta: new Int8Array(1)};
  delete obj2?.ta[0];             // jsse: true, no throw -- should throw TypeError
})();
```

Verified against `target/release/jsse` (built from HEAD `077e4ec3`):

```
$ ./target/release/jsse -e "var obj = {ta: new Int8Array(1)}; console.log(delete obj?.ta[0])"
true
```

Node v26.9.0: `false`.

## Fix shape

[...]

PR #660, opened 2026-09-20
fix(optional-chain): throw on private access with non-object base

### LINKED ISSUE #642: bug: optional-chain private field access on non-object silently returns undefined instead of throwing
## Summary

`obj?.value.#x` — a private-field access reached through an optional chain — silently evaluates to `undefined` when the base of the private access is a non-object primitive, instead of throwing `TypeError` as the spec requires. The same private-field access performed *outside* an optional chain throws correctly.

## Spec

PrivateFieldGet (ECMA-262 [`MemberExpression : MemberExpression . PrivateIdentifier`](https://tc39.es/ecma262/#sec-property-accessors-runtime-semantics-evaluation) / `PrivateGet`) requires the base to be an Object; if not, a `TypeError` is thrown.

## Root cause

- `src/interpreter/eval.rs:3940` (`private_get`, used for a private access *not* part of an optional chain) explicitly throws `TypeError: Cannot read private member #x from a non-object` when the base isn't an object.
- `src/interpreter/eval/access.rs:277-309` (the optional-chain tail evaluator, `eval_oc_tail_with_this_ctx`) has its own private-field arm that instead falls through to `return Ok((JsValue::UNDEFINED, inner_val))` at line ~306-308 when the base is a non-object primitive reached via a chain — silently swallowing the error.

## Repro

```js
class C {
  #x = 1;
  direct(v)  { return v.#x; }        // no optional chain anywhere
  chained(o) { return o?.value.#x; } // private access is a non-optional link inside a chain
}
const c = new C();
c.direct(5);            // throws correctly in jsse
c.chained({value: 5});  // jsse: returns undefined -- should throw
```

```
$ ./target/release/jsse repro.js
undefined
```

Node v26.9.0:

```
TypeError: Cannot read private member #x from an object whose class did not declare it
```

## Fix shape

`eval_oc_tail_with_this_ctx`'s private-field arm in `src/interpreter/eval/access.rs` needs the same non-object check and `TypeError` throw as `private_get` in `src/interpreter/eval.rs:3940`.

## Related

Same root-cause family (optional-chain tail evaluator in `access.rs` missing special-case handling present in the corresponding `eval.rs` path) as the `delete obj?.ta[0]` bug — see the linked issue once filed.

---
Found via a `slopo` semantic-duplication scan (cluster 158) cross-referencing `eval.rs`'s and `access.rs`'s private-field-access code paths, then confirmed by hand against Node.

pmatos:
## Summary

`o?.value.#x` and `o.#x?.y` silently evaluated to `undefined` when the private-name base was a non-object, instead of throwing `TypeError` (ECMA-262 §6.2.5.5 GetValue 3.a/3.b → §7.3.28 PrivateGet). `o.#x?.y` also returned `undefined` for a set-only private accessor. Both optional-chain arms in `src/interpreter/eval/access.rs` re-implemented the private-element lookup inline and dropped the `TypeError` arms of `Interpreter::private_get`; they now delegate to it.

## Beyond the issue's fix shape

The `eval_oc_base` private arm applied the chain tail itself, and every caller (`eval_optional_chain_with_ref`, the `eval.rs` OptionalChain arm, the delete path) applies it again. A non-nullish `o.#x?.y` therefore evaluated `.y` twice (`#x = {y: "found"}` returned `undefined`; Node returns `"found"`). The arm now returns `(value, base)` like the `Dot`/`Computed` arms, so the chain is applied once and `this` is preserved for `o.#m?.()`. This removed the now-unused `chain` parameter of `eval_oc_base` and the dead `eval_oc_tail_with_this` helper.

Out of scope, unchanged: `super.#x` in an optional-chain base (a parser-level SyntaxError, not a runtime one) and `delete` via optional chain (#643).

## Testing

- New `test262-extra/optional-chain-private-field-non-object-base.js` (esid `sec-privateget`): tail-of-chain and base-of-chain reads on primitive / nullish / brand-less bases, field / method / getter / set-only accessor, `?.()` receiver. Failed before the change (both modes), passes after; also passes on Node.
- Full test262: 99,911/99,911, 0 regressions. `test262-extra/`: 338/338. `run-custom-tests.py`: 11/11. `cargo test --release`: all green. `./scripts/lint.sh`: green.
- Note: `cargo clippy --tests -D warnings` reports two `cloned_ref_to_slice_refs` lints in `src/interpreter/builtins/iterators.rs` (test code, untouched here; newer toolchain lint). `lint.sh` does not build tests, so the gate is unaffected.

Closes #642


pmatos:
Automated \`/code-review --fix\` pass applied fixes (d6ba48f):

- Optional-chain call tail: root callee/\`this\` across argument evaluation; computed-key arms root the base across key evaluation (GC could previously free them, e.g. \`o?.#g(\$262.gc())\`).
- Route the remaining hand-inlined private reads (\`eval_oc_base\` super branch, \`eval_call\` private callee arm) through \`private_get\`.
- test262-extra: corrected spec step reference/feature tags, added coverage cases, new \`optional-chain-call-and-key-gc-rooting.js\`.

Gate: lint, release build, \`cargo test --release\` clean; full test262 99,911/99,911 pass, 0 regressions.

Pre-existing issues found but not fixed (out of scope, need parser/generator-transform changes): \`?.()\` containing yield/await loses receiver in generator_transform; \`delete o?.#x\` is not an early SyntaxError; private-name tagged template tag always throws.

Issue #663, opened 2026-09-20
bug: if/while/for condition throw inside a generator's try block bypasses catch/finally when the body contains a yield


pmatos:
## Summary

Sibling of #646. A generator's `StateTerminator::ConditionalGoto` (an `if`/`while`/`for` whose condition contains no `yield` but whose body does) evaluates its condition with a bare `match`, and on `Completion::Throw(e)` finishes the generator (sync) or rejects the request promise (async) instead of calling `route_exception!` → `route_generator_exception`. An enclosing `catch`/`finally` never runs.

Sites in `src/interpreter/eval/generator_runtime.rs`: `ConditionalGoto` in `generator_next_state_machine_impl` (~1330) and in `async_generator_next_state_machine_impl` (~5170).

## Repro

```js
function thrower() { throw new Error('boom'); }
function* a(){ try { if (thrower()) { yield 1; } } catch(e){ yield 'caught-if:' + e.message; } }
function* b(){ try { while (thrower()) { yield 1; } } catch(e){ yield 'caught-while:' + e.message; } }
function* c(){ try { for (;thrower();) { yield 1; } } catch(e){ yield 'caught-for:' + e.message; } }
function* d(){ try { if (thrower()) { yield 1; } } finally { console.log('fin-if'); } }
for (const f of [a,b,c]) { try { console.log(JSON.stringify(f().next())); } catch (e) { console.log(f.name, 'ESCAPED', e.message); } }
try { d().next(); } catch (e) { console.log('d escaped', e.message); }
async function* aa(){ try { if (thrower()) { yield 1; } } catch(e){ yield 'caught-async-if:' + e.message; } }
aa().next().then(r=>console.log('resolved', JSON.stringify(r)), e=>console.log('rejected', e.message));
```

jsse: `a`/`b`/`c` ESCAPED boom, `d` escaped without printing `fin-if`, async `rejected boom`.
Node: `caught-if:boom`, `caught-while:boom`, `caught-for:boom`, `fin-if` then `d escaped boom`, async `resolved {"value":"caught-async-if:boom","done":false}`.

## Fix shape

Same as #646: route the `Completion::Throw` through `route_exception!` (at the driver-loop level, not inside a nested `for`) before the existing "complete generator / reject promise" cleanup. Other terminators with the same bare-throw pattern should be audited in the same pass.

Found while implementing #646 (slice 7 sibling probe).

PR #664, opened 2026-09-20
fix(generators): route switch discriminant/case-test throws through enclosing try

### LINKED ISSUE #646: bug: switch inside a generator's try block bypasses catch/finally when discriminant or case-test throws
## Summary

When a `switch` statement appears inside a `try` block inside a generator (sync or async) **and** the switch body contains a `yield`, an exception thrown while evaluating the switch discriminant or a `case` test does not propagate through the enclosing `try`'s `catch`/`finally` — it escapes the generator entirely instead. A `switch` with no `yield` inside it, or a switch outside a generator, is unaffected (it executes as a single ordinary AST-walk step, not through the generator state machine).

## Spec

SwitchStatement Evaluation (ECMA-262 §14.12.3): an abrupt completion (throw) from evaluating the discriminant or a `CaseClause` test must propagate like any other abrupt completion, i.e. through any enclosing `try`/`catch`/`finally`.

## Root cause

`src/interpreter/eval/generator_runtime.rs`:
- Sync generator: discriminant eval ~1440-1454, case-test eval ~1457-1478 (inside `generator_next_state_machine_impl`).
- Async generator: discriminant eval ~5324-5347, case-test eval ~5350-5374 (inside `async_generator_next_state_machine_impl`).

Every other abrupt-completion site in these same state-machine drivers (e.g. `ForOfInit`'s iterable/iterator evaluation, right next to this code at ~1499-1528) routes the exception through `route_exception!` → `route_generator_exception`, which walks `try_stack` to find an active `catch`/`finally` handler and resumes state-machine execution there. The switch-discriminant/case-test code does not call this — it evaluates with a bare `match` and on `Completion::Throw` returns `Completion::Throw(e)` directly (sync) or rejects the promise (async), bypassing `try_stack` entirely.

## Repro

```js
function thrower() { throw new Error('boom'); }
function* g(){
  try { switch (0) { case thrower(): yield 'z'; break; } }
  catch(e){ yield 'caught:' + e.message; }
  yield 'after';
}
console.log(JSON.stringify(g().next()));
```

```
$ ./target/release/jsse repro.js
Error: boom
$ echo $?
1
```

Node v26.9.0: `{"value":"caught:boom","done":false}`.

`finally` is also skipped (resource-cleanup symptom):

```js
function* g(){
  try { switch (thrower()) { case 1: yield 'z'; break; } }
  finally { console.log('cleanup-ran'); }
  yield 'after';
}
try { g().next(); } catch (e) { console.log('escaped: ' + e.message); }
```

jsse only prints `escaped: boom` (`cleanup-ran` never runs). Node prints `cleanup-ran` then `escaped: boom`.

Async generators show the same bug (promise rejects instead of resolving through `catch`):

```js
async function* ag(){
  try { switch (thrower()) { case 1: yield 'z'; break; } }
  catch(e){ yield 'caught:' + e.message; }
}
ag().next().then(r => console.log('resolved:'+JSON.stringify(r)), e => console.log('rejected:'+e.message));
```

jsse: `rejected: boom`. Node: `resolved:{"value":"caught:boom","done":false}`.

## test262 coverage gap

`uv run python scripts/run-test262.py test262/test/language/statements/generators/` is 510/510 (100%) on current HEAD — this specific intersection (switch discriminant/case-test throw, inside try, inside a yield-containing switch, inside a generator) is not covered by test262. Per this repo's CLAUDE.md convention for spec-correct-but-untested-by-test262 gaps, this should get a regression test in `test262-extra/` once fixed.

## Fix shape

Route the discriminant and case-test evaluation in both `generator_next_state_machine_impl` and `async_generator_next_state_machine_impl` through the same `route_exception!`/`route_generator_exception` mechanism used by the neighboring `ForOfInit` code, instead of returning/rejecting directly on `Completion::Throw`.

---
Found via a `slopo` semantic-duplication scan (cluster 413) comparing the switch-dispatch code against the neighboring `ForOfInit` exception-routing code in the same state-machine driver, then confirmed by hand against Node (and independently re-verified before filing).

pmatos:
## Summary

In a sync or async generator, a `switch` whose body contains a `yield` is lowered to `StateTerminator::SwitchDispatch`. The state-machine drivers evaluated the discriminant and each case test with a bare `match` and, on `Completion::Throw`, finished the generator (sync) or rejected the request promise (async). The throw never reached `route_generator_exception`, so an enclosing `catch`/`finally` never ran (ECMA-262 §14.12.4 / §14.12.3 / §14.15.3).

Both `SwitchDispatch` arms (`generator_next_state_machine_impl`, `async_generator_next_state_machine_impl` in `src/interpreter/eval/generator_runtime.rs`) now evaluate inside a labeled block that yields `Result<usize, JsValue>`, and call `route_exception!` once after it. Routing outside the case `for` is deliberate: the macro expands to `continue`, which would otherwise target the inner loop and resume from the wrong state. The unhandled tail (complete the generator, return `Throw` / reject the promise) is unchanged, and so is the different "other completion" handling in the sync and async arms.

## Behaviour note

With no enclosing `try`, a throwing switch dispatch inside a `for-of` body now closes the loop's iterator (`return()` called once) before the throw escapes. That is what ForIn/OfBodyEvaluation requires and matches the existing `ForOfInit` routing. It was previously skipped. The new tests pin it.

## Tests

- `test262-extra/generator-switch-abrupt-completions-through-try.js` and `test262-extra/async-generator-switch-abrupt-completions-through-try.js`: discriminant and case-test throws through `catch`, `finally`, nested `catch`+`finally`, unhandled (generator completed afterwards), case-test evaluation order (later tests not evaluated), and for-of unwinding with and without a `try`. Both failed before the fix.
- The three repros from the issue now match Node.
- `./scripts/lint.sh` clean; `cargo test --release` green.
- Full `run-test262.py -j 32`: 99911/99911, 0 regressions. `test262-extra/`: 371/371. `run-custom-tests.py`: 11/11.
- `test262-pass.txt` not touched.

## Follow-up

`ConditionalGoto` (an `if`/`while`/`for` condition with no `yield` but a `yield` in the body) has the same bypass in both drivers. It reproduces and is filed as #663 rather than bundled here.

Closes #646


pmatos:
Automated `/code-review --fix` pass on this PR.

**Fixed and pushed (5b66b79):**
- `SwitchDispatch` unhandled-throw arms (sync + async) now call `dispose_resources`, matching the body-throw paths (`using` cleanup was skipped).
- Async `SwitchDispatch` no longer swallows `Completion::Exit` from `__host_exit` in a discriminant or case test.
- Regression tests: two unit tests in `src/interpreter/tests.rs`, `using` disposal cases in both `test262-extra/*generator-switch-abrupt-completions-through-try.js` files.

Gate after the fix: `./scripts/lint.sh`, `cargo build --release`, `cargo test --release`, full test262 (99,911/99,911, 0 regressions), and test262-extra (371/371) are all green.

**Found but deliberately not fixed (separate defects, outside this diff; worth follow-up issues):**
- `ConditionalGoto` terminator (sync ~L1330, async ~L5167) finishes the generator on a throw instead of routing it through the enclosing try, so `if`/`while`/`for`/`do-while` conditions that throw inside `try` bypass `catch`.
- Async `Return` terminator (~L4762) does not route a throwing return expression through the try stack.
- `case (yield x):` is not decomposed by `generator_transform.rs` (pre-existing).
- `route_generator_exception` does not clear `pending_return` when a throw supersedes an in-progress `return()`.
- Altitude: each terminator hand-rolls route/dispose/mark-completed; a shared "terminator threw" helper per driver would prevent this class of drift.

pmatos:
Simplify pass (pushed e80cbfc): reused `discard_generator_for_of_loops_on_exit` for the async switch-dispatch Exit arm, dropped the dead `unreachable!` arm, compacted the Yield fallback, and merged the two host-exit tests. Full gate green (lint, release build, cargo test, test262 99,911/99,911, 0 regressions).

Skipped as out of scope for #646 (behaviour changes or larger refactors): a shared switch-target helper or a `term_eval!` macro across terminators, and per-iteration terminator cloning (pre-existing).

Separate bug found, not fixed here: `ConditionalGoto` in generator_runtime.rs (sync ~1330, async ~5167) still doesn't route a throwing condition through the enclosing try. `function* g(){ try { yield 1; if (t()) { yield 5 } } catch(e){ yield 'caught' } }` escapes the catch on jsse, while node yields 'caught'. `while`, `for` and ternary conditions likely share it. Worth its own issue.

PR #667, opened 2026-09-20
fix(parser): reject strict-reserved name on async function expression

### LINKED ISSUE #647: bug: async function expression skips strict-mode-reserved-word early error for its name
## Summary

A named `async function` (or `async function*`) *expression* whose name is a strict-mode-reserved word (`implements`, `interface`, `package`, `private`, `protected`, `public`, `static`, `let`, `yield`) should be a `SyntaxError` in strict-mode code, per the same early-error rule applied to plain function expressions and to function/async-function *declarations*. jsse enforces this for the sibling forms but not for async function expressions.

## Spec

BindingIdentifier early error (ECMA-262 §15.1.1) for a named function/async-function expression in strict-mode code: the name must not be `eval`/`arguments` nor a strict-mode-reserved word.

## Root cause

`src/parser/expressions.rs:2159-2168` — `parse_async_function_expression`'s `body_strict` block is missing the `Self::is_strict_reserved_word(n)` check. The equivalent checks exist at:
- `src/parser/expressions.rs:2106` (`parse_function_expression`)
- `src/parser/declarations.rs:285` (`parse_function_declaration`, which already covers both sync and async *declarations* via its `is_async` flag — only the *expression* form is affected)

Since `is_generator` is read inside the same function, this also covers `async function*` expressions.

## Repro

```js
(function package(){"use strict";});        // throws (correct)
(async function package(){"use strict";});  // should throw, does not
(async function* package(){"use strict";}); // should throw, does not
```

```
$ ./target/release/jsse -e '(async function package(){"use strict";}); console.log("no throw")'
no throw
```

Node v26.9.0: `SyntaxError: Unexpected strict mode reserved word`.

## Fix shape

Add the same `Self::is_strict_reserved_word(n)` check inside `parse_async_function_expression`'s `body_strict` block in `src/parser/expressions.rs`, mirroring `parse_function_expression`.

---
Found via a `slopo` semantic-duplication scan (cluster 562) comparing the strict-mode name-validation wiring across function-expression parsing sites, then confirmed by hand against Node.

pmatos:
`parse_async_function_expression` only checked `eval`/`arguments` in its post-body `body_strict` block, so a named async function or async generator expression whose own body has a `"use strict"` directive accepted a strict-mode reserved word as its name:

```js
(async function package(){"use strict";});   // no throw before, SyntaxError now
(async function* package(){"use strict";});  // no throw before, SyntaxError now
```

The fix adds the same `Self::is_strict_reserved_word(n)` check that `parse_function_expression` and `parse_function_declaration` already have. `async function*` shares that block, so it needs no separate handling.

Spec: Identifiers, Static Semantics: Early Errors (`sec-identifiers-static-semantics-early-errors`), together with Strict Mode Code (`sec-strict-mode-code`). A `"use strict"` directive in the body makes the whole function, including its `BindingIdentifier`, strict.

## Tests

- New `test262-extra/async-function-expression-name-strict-reserved-word.js` covers all nine reserved words for the async and async-generator expression forms. It also covers sibling forms and outer-strict inheritance as regression coverage, `eval`/`arguments`, and sloppy or non-reserved controls that must still parse. test262 has no test for this case.
- Before the fix the file failed in both strict and sloppy scenarios. `assert.throws` aborts on the first failure, so only `(async function implements(){'use strict';})` was observed failing; the async-generator repro is documented in the issue.
- `./scripts/lint.sh`, `cargo test --release`, and `run-custom-tests.py` (11/11) are clean.
- Full test262: 99,911/99,911 scenarios pass, 0 regressions. `test262-extra/`: 373/373 pass.
- `test262-pass.txt` is untouched.

Closes #647


pmatos:
Automated `/code-review --fix` pass on #667 (pushed as 779f532).

**Applied:** extracted `Parser::check_strict_function_name` (the three copy-pasted checks had drifted, which caused #647) and tightened the test262-extra test (`flags: [noStrict]`, correct spec quote, duplicate removed, escaped-name/directive-position/sloppy negative cases added).

**Gate on this push:** `./scripts/lint.sh`, `cargo build --release`, `cargo test --release` all pass; full `run-test262.py -j 32` is 99911/99911 with 0 regressions; `test262-extra/` is 372/372.

**Not fixed (pre-existing, outside this diff; each deserves its own issue):**
- A parenthesized `('use strict')` statement is treated as a directive (`is_string_literal_statement` has no paren tracking), so `(async function let(){ ('use strict'); })` now throws where it previously parsed. Needs AST paren tracking.
- Arrow functions don't re-check parameter names against their own `'use strict'` body (`(let) => {'use strict'}`).
- Async/generator/async-generator object methods skip `check_strict_params` on a `'use strict'` body (`({ async m(let){'use strict'} })`).
- `check_duplicate_params_strict` runs unconditionally for async/generator functions, rejecting sloppy `(async function f(a,a){})`.
- `class eval {}` / `class arguments {}` accepted; `async => 1` / `of => {}` rejected; wrong `[Yield]` context for method/async-arrow params inside generators.

Issue #671, opened 2026-09-20
bug: yield inside a switch case test expression throws TypeError: Iterator next failed


pmatos:
Found while fixing #651. `SwitchDispatch` in `src/interpreter/eval/generator_runtime.rs` evaluates case test expressions with a plain `eval_expr`, so a `yield` there is not a suspension point.

```js
function* h() { switch (1) { case (yield 5, 1): return 'x'; } }
console.log(JSON.stringify([...h()]));
// node: [5]    jsse: TypeError: Iterator next failed
```

Spec: `sec-runtime-semantics-caseblockevaluation` (case selectors are evaluated in order, and `yield` is a valid AssignmentExpression there).

Issue #670, opened 2026-09-20
bug: for-in with yield in a generator body yields nothing


pmatos:
Found while fixing #651, at `181dbe9`. A `for-in` whose body contains `yield` completes immediately.

```js
function* g(o) { for (var k in o) { yield k; } }
console.log(JSON.stringify([...g({a: 1, b: 2})]));
// node: ["a","b"]    jsse: []
```

Independent of `switch`. Probably a regression in the for-in state-machine lowering — worth bisecting.

Spec: `sec-for-in-and-for-of-statements-runtime-semantics-labelledevaluation`, `sec-generatorresume`.

Issue #669, opened 2026-09-20
bug: yield-free try/with containing break/continue in a switch case of a generator still falls through


pmatos:
Found while fixing #651. `stmt_has_break_or_continue` (`src/interpreter/generator_transform.rs`) does not recurse into `Try` / `With` (nor into nested loops/switches for labeled or outer jumps), so a yield-free clause whose `break`/`continue` sits inside a `try` is still emitted verbatim and the raw completion is dropped by the state driver.

```js
function* t(x) {
  var l = [];
  switch (x) {
    case 1: try { break; } finally { l.push('f'); }
    case 2: l.push('two'); break;
    case 3: yield 0;
  }
  yield l.join();
}
console.log(JSON.stringify([...t(1)]));
// node: ["f"]    jsse: ["f,two"]
```

A prototype that only recursed into `Try`/`With` was still wrong (finalizer skipped, `with` scope), so this needs its own design: `try`-depth-aware `LoopControl` targets and `with_scopes`. The same shape affects loops, e.g. `for (...) { try { if (c) break; } finally {} yield i; }`.

Spec: `sec-runtime-semantics-caseblockevaluation`, `sec-try-statement-runtime-semantics-evaluation`.

pmatos:
Planning notes for #669 (autonomous run, judgement calls recorded so a reviewer can override).

**Reproduced** on a release build of this branch: the issue example gives `["f,two"]` (node: `["f"]`). The same dropped-jump happens for loops (`for { try { if (c) break; } finally {} yield i; }`), for `with` (`case 1: with (o) { ...; break; }`), and in `async function` / `async function*`.

**Root cause:** the yield-free `try`/`with` runs natively in the tree-walker, so its finalizer and `with` scope are already correct. Only the resulting raw `Completion::Break`/`Continue` is dropped, because none of the three state drivers inspects it.

**Chosen approach (differs from the issue's suggestion):** instead of lowering `try`/`with` into states, keep them inline and record, per state, the jumps that escape the emitted statement (label-aware, resolved against `break_targets`/`continue_targets` at emit time). The drivers substitute that recorded `Goto`/`LoopControl` terminator when a raw `Break`/`Continue` surfaces. This avoids re-implementing finalizer/catch semantics in states and avoids the "finalizer skipped" failure of a lowering prototype.

[...]

Issue #53, opened 2026-04-14
js-tokens benchmark: generator-based tokenizer produces wrong token count


pmatos:
## Description

The JetStream 3 js-tokens benchmark uses a generator-based JavaScript tokenizer. jsse produces 132 tokens for a JSX source snippet where Node produces 126. The 6 extra tokens per iteration accumulate to an invalid total count.

```
Error: this.tokenCount of 115475 is invalid!
```

## Analysis

The first JSX `<div className="Comment">` tag's opening `<` is tokenized as a regular `Punctuator` instead of `JSXPunctuator`, and `div` becomes `IdentifierName` instead of `JSXIdentifier`. This means the tokenizer's JSX mode entry is not being triggered correctly, causing different token boundaries and 6 extra tokens per sample.

This is likely a semantic difference in how the generator-based tokenizer's state machine interacts with the engine's generator replay mechanism — the generator state after yield may not be perfectly restored, causing the tokenizer to lose track of its JSX parsing mode.

## Affected JetStream 3 benchmarks

- js-tokens

PR #672, opened 2026-09-20
fix(generators): lower yield-free break/continue in switch cases of state-machine bodies

### LINKED ISSUE #651: bug: JetStream js-tokens still fails with "this.tokenCount of 115475 is invalid" (same symptom as closed #53)
## Description

The JetStream `js-tokens` workload fails with:

```
Error: this.tokenCount of 115475 is invalid!
```

This is the same symptom, with the same count (115475), as #53, which was closed as completed by #76. Either that fix did not cover this case or the JetStream revision differs from the one #53 was filed against. Per #53's analysis the generator-based tokenizer produced 132 tokens for a JSX snippet where Node produces 126 (JSX mode entry not triggered). Runs for ~21 s before failing (the 2026-03-17 build timed out at 120 s, so it never got this far). Same result with `--bytecode`.

## Reproduction

```bash
# JetStream, pinned to the revision used in docs/perf reports; assets ship compressed
git clone --depth 1 https://github.com/WebKit/JetStream /tmp/JetStream && cd /tmp/JetStream
git fetch --depth 1 origin c603c04db8505477867974a69789309ded2cc948 && git checkout c603c04db8505477867974a69789309ded2cc948
npm install && npm run decompress -- --quiet     # needs node >= 22
# from the jsse checkout
cargo build --release
uv run python scripts/run-jetstream.py --test js-tokens --iterations 1 --timeout 120 --engine target/release/jsse --jetstream /tmp/JetStream
```

To compare with Node, run the same workload file under `node` and diff the token stream, as in #53.

## Environment

jsse v0.8.1 (`083746f`, release build), buildbox3 (AMD EPYC 7501, Linux 6.1), JetStream `c603c04`, `--iterations 1 --timeout 120 -j 1`, N=3 repeats. Found while collecting data for the 2026-09-18 comparison against the 2026-03-17 build (`7a4d095`).

pmatos:
## Summary

The JetStream `js-tokens` failure (`this.tokenCount of 115475 is invalid`) is a control-flow bug in the generator/async state-machine transform, not generator replay or JSX-mode state.

`transform_switch_statement` (`src/interpreter/generator_transform.rs`) splits a `switch` that contains a suspension into one state per clause, chained by `Goto(next_clause)`. A clause **without** a suspension was emitted verbatim, so its `break;` / `continue;` ran natively and surfaced as a raw `Completion::Break` that the state driver has no target for and ignores. Control then continued to `Goto(next_clause)` — the clause **fell through** (CaseBlockEvaluation stops at the first abrupt completion, `sec-runtime-semantics-caseblockevaluation`). Clauses that did contain a suspension went through `transform_statements`, which lowers `break` to a `Goto`, which is why this was masked in most generator tests.

In `js-tokens`, the punctuator `switch` has yield-free `case "(":` / `")"` / `"{"` ahead of yielding `case "}"` / `case "<"`. `(` fell into `case "}"` and corrupted `lastSignificantToken`, so the `<` after `return (` was no longer recognised as JSX entry: 132 tokens for the JSX sample instead of Node's 126, i.e. 250 × 6 = 1500 extra tokens (115475 vs 113975).

**Fix:** route a clause body through `transform_statements` when any statement has a suspension **or** `stmt_has_break_or_continue`; yield-free, jump-free bodies keep the verbatim path (so the async-generator `return` / `await using` special-casing in `transform_statements` is not newly applied to them).

Note on history: the diagnosis in #53 (generator replay / JSX mode) was wrong and #76 (var hoisting) was unrelated to this.

## Verification

- New `test262-extra/` tests, red before the fix and green after (also cross-checked against Node):
  - `generator-switch-yield-free-case-break-does-not-fall-through.js` (incl. a positive fall-through guard)
  - `generator-switch-yield-free-case-continue-and-labeled-break.js`
  - `async-switch-yield-free-case-break-does-not-fall-through.js` (async function, async generator, for-await)
  - `generator-switch-punctuator-dispatch-yield-free-cases-before-yielding-case.js` (reduced `js-tokens` shape)
- Full `uv run python scripts/run-test262.py -j 32`: 99,911 / 99,911 scenarios, 0 regressions.
- `test262-extra/` 389/389, `run-custom-tests.py` 11/11, `cargo test --release` green, `./scripts/lint.sh` green.
- JetStream `c603c04`, `--test js-tokens,lazy-collections --iterations 1 --repeats 3 --no-idle-gate`: `PASS js-tokens`, `PASS lazy-collections` (idle gate disabled on a loaded host, so timing is diagnostic only; this checks correctness).

## Out of scope (filed separately)

- #669 — yield-free `try`/`with` containing `break`/`continue` in a switch clause still falls through (`stmt_has_break_or_continue` doesn't recurse into `Try`/`With`)
- #670 — `for (var k in o) { yield k }` in a generator yields nothing
- #671 — `yield` in a `case` test expression throws `TypeError: Iterator next failed`

Closes #651


pmatos:
## Automated code-review pass (`/code-review --fix`)

**Fixed and pushed (259082e):** an infinite loop introduced by this PR. When a switch is the last statement of a loop body, `transform_switch_statement`'s unconditional `Goto(next_state)` overwrote the already-finalized loop test/head state, so the loop spun forever. It is now skipped when `current_state_id == next_state`, as the while and if transforms do. Regression test: `test262-extra/generator-switch-last-case-tail-break-as-loop-body-tail.js`.

Quality gate on the fixed tree: `./scripts/lint.sh` clean, `cargo build --release` OK, `cargo test --release` green, full test262 `-j 32` at 99911/99911 with 0 regressions. JetStream `js-tokens` still gives 113975 (valid).

**Not fixed — follow-up needed.** These are regressions or gaps versus node/the pre-PR build; test262 does not catch any of them. Most fixes belong in the drivers (`generator_runtime.rs` and the async-function driver in `eval.rs`), which should map a raw `Completion::Break`/`Continue` to the recorded switch or loop target rather than in the `stmt_has_break_or_continue` predicate.

1. **Block scoping lost.** Yield-free clause bodies with break/continue go through `transform_statements`, which flattens `{ let/const/class ... }` blocks into the function env. `for(let i=0;i<3;i++){ switch(i){ case 5: yield 1; default:{ let v='d'+i; fns.push(()=>v); break; } } }` closes over one `v` (`def2,def2,def2`; node `def0,def1,def2`). Shadowing leaks the same way.
2. **`using` not disposed** in a flattened case block (`body,after`; node `body,disposed,after`).
3. **Block-level function declarations** in a flattened case block are not hoisted (`ReferenceError`), and leak to function scope in strict mode.
4. **`finally` bypassed.** Lowered break/continue uses `Goto` in sync and async generators (`LoopControl` only for async functions), so enclosing `finally` blocks are skipped and a stale try_stack entry can remain.
5. **`stmt_has_break_or_continue` is incomplete.** It does not recurse through `Try`, `With`, nested loops (labeled targets) or nested switch (`continue`), and `transform_labeled_statement` emits yield-free labeled bodies verbatim. Those clauses still fall through — the same symptom as #651. Its comment ("Don't recurse into nested loops/switch") is inaccurate for labeled targets.
6. **Async-generator `return` timing** differs between `case 0: return 1;` and `case 0: if(k) break; return 1;`.
7. **Test gaps.** The new tests have no let/const/class/function/`using` declarations in a clause with break, no break through try/with, and no nested loop/switch escape. The async file largely duplicates the generator tests.

Pre-existing (not from this PR): a generator `for-in` containing `yield` yields nothing; `while(i<3){ i++; lbl: if(i){ yield i; } }` hangs because `transform_labeled_statement` has the same unconditional-finalize flaw.

pmatos:
Simplify pass: no source changes. Reuse, efficiency and altitude were clean; the one simplification idea (drop the case-body pre-filter and call `transform_statements` unconditionally) would change how yield-free async `return` / `await using` cases are lowered, so I skipped it. Only change pushed: a stale sentence in a `test262-extra` description (d2c8ee4). Lint, `cargo test --release` and full test262 (99,911/99,911, 0 regressions) are green.

Separate pre-existing bug, not fixed here (out of scope for a quality pass): an `await using` block ending in `break` inside a switch case of an async generator still falls through.

```js
async function* g(x) {
  switch (x) {
    case 2: { await using r = { async [Symbol.asyncDispose]() {} }; log.push("two"); break; }
    default: log.push("def"); yield 9;
  }
}
```
Expected log `two`; jsse gives `two,def,9`. The same shape with a plain `let` block is correct, so the gap is in the `await using` path of `transform_yielding_statement`.

